PR #27, opened 2021-12-23
Define EventSourceSubscription2021 type for Solid Notifications

acoburn:
The Solid Notification Protocol refers to an `EventSourceSubscription2021` type but there is no detailed description of that type. This document defines the requirements for `EventSourceSubscription2021`.


eventsource-subscription-2021.html
```diff
@@ -0,0 +1,499 @@
...
+
+              <p property="skos:definition">The <code>EventSourceSubscription2021</code> type defines the following properties:</p>
+
+              <span rel="skos:hasTopConcept"><span resource="#eventsourcesubscription2021-endpoint"></span></span>
+
+              <dl>
+                <dt about="#eventsourcesubscription2021-endpoint" property="skos:prefLabel" typeof="skos:Concept"><dfn id="eventsourcesubscription2021-endpoint">endpoint</dfn></dt>
+                <dd about="#eventsourcesubscription2021-endpoint" property="skos:definition"><span about="" id="eventsourcesubscription2021-endpoint-https" rel="spec:requirement" resource="#eventsourcesubscription2021-endpoint-https"><span property="spec:statement">The <code>endpoint</code> property is used in the body of the subscription response. The value of this property <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> be a URI, using the https scheme.</span></span> A JavaScript client would use the entire value as input to the EventSource constructor.</dd>
```
csarven, line 384:
```suggestion
                <dd about="#eventsourcesubscription2021-endpoint" property="skos:definition"><span about="" id="eventsourcesubscription2021-endpoint-https" rel="spec:requirement" resource="#eventsourcesubscription2021-endpoint-https"><span property="spec:statement">The <code>endpoint</code> property is used in the body of the subscription response. The value of this property <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> be a URI, using the <code>https</code> scheme.</span></span> A JavaScript client would use the entire value as input to the EventSource constructor.</dd>
```

csarven (commented):
If the general idea / best practice for servers to allow only one type of subscription per endpoint, this should be mentioned. AFAIK, EventSource doesn't have a parameter along the lines of a subprotocol like in WebSocket, so it can't indicate which type is used. This may entail that a given endpoint can/should? only accept one type or a type that's backwards compatible.

elf-pavlik (approved):

PR #29, opened 2022-01-03
Replace separate gateway endpoint with RDF-based protocol discovery

### LINKED ISSUE #24: Reconsider necessity of gateway API in notification protocol
At present, the Notification Protocol defines a [gateway API](https://solid.github.io/notifications/protocol#notification-gateway) that acts as a discovery endpoint for clients: agents submit `POST` JSON-LD requests and receive JSON-LD response bodies. This is an unauthenticated endpoint.

Rather than providing a required request-response gateway API, it would arguably be simpler do just define the subscription types that are supported in a single RDF document. Clients can fetch that RDF document and, given the features supported by the various subscription types, the client can simply choose which path to follow.

One can argue that the current gateway API model already has such an RDF document implicitly and the HTTP API brokers access to that document. Removing that API has several advantages:

* The server implementation provides what is, effectively, a static document, thereby making the implementation considerably simpler
* Any client code needs only to fetch and parse a single resource and can then decide on its own what subscription type(s) it uses, which arguably means fewer request/response cycles for the client

acoburn:
Resolves #24

This entirely removes the Gateway endpoint, replacing it with an RDF-based resource that can be used by clients to find the appropriate subscription API, based on the features and protocols supported by the client.

For example, the following RDF structure could replace the Gateway endpoint:

```turtle
@prefix solid: <http://www.w3.org/ns/solid/terms#> .
@prefix notify: <http://www.w3.org/ns/solid/notifications#> .

<>
    a solid:StorageMetadata ;
    notify:hasWebSocketNotification <#websocketNotification> ;
    notify:hasWebHookNotification <#webhookNotification> ;
    notify:hasLinkedDataNotification <https://ldn.example/Metadata> .
<#websocketNotification>
    a notify:WebSocketSubscription2021 ;
    notify:endpoint <https://websocket.example/subscription> ;
    notify:features notify:state, notify:rate, notify:expiration .
<#webhookNotification>
    a notify:WebHookSubscription2021 ;
    notify:endpoint <https://webhook.example/subscription> ;
    notify:features notify:rate, notify:expiration, notify:webookAuth .
```

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -457,17 +457,17 @@ content: "";
                     </thead>
                     <tbody>
                       <tr>
-                        <td>rdf</td>
+                        <td><code>rdf</code></td>
                         <td>http://www.w3.org/1999/02/22-rdf-syntax-ns#</td>
                         <td>[<cite><a class="bibref" href="#bib-rdf-schema">rdf-schema</a></cite>]</td>
                       </tr>
                       <tr>
-                        <td>solid</td>
+                        <td><code>solid</code></td>
                         <td>http://www.w3.org/ns/solid/terms#</td>
                         <td>Solid Terms</td>
                       </tr>
                       <tr>
-                        <td>notify</td>
+                        <td><code>notify</code></td>
                         <td>http://www.w3.org/ns/solid/notifications#</td>
                         <td>Solid Notifications</td>
                       </tr>
@@ -522,20 +522,18 @@ content: "";
 <code>@prefix notify: &lt;http://www.w3.org/ns/solid/notifications#&gt; .</code>
 <code></code>
 <code>&lt;&gt;</code>
-<code>    a solid:StorageMetadata ;</code>
-<code>    notify:hasNotificationChannel &lt;#websocketNotification&gt; , &lt;#webhookNotification&gt;</code>
-<code>.</code>
+<code>  a solid:StorageMetadata ;</code>
+<code>  notify:hasNotificationChannel &lt;#websocketNotification&gt;, &lt;#webhookNotification&gt; .</code>
+<code></code>
 <code>&lt;#websocketNotification&gt;</code>
-<code>    a notify:WebSocketSubscription2021 ;</code>
-<code>    notify:endpoint &lt;https://websocket.example/subscription&gt; ;</code>
-<code>    notify:features notify:state, notify:rate, notify:expiration</code>
-<code>.</code>
+<code>  a notify:WebSocketSubscription2021 ;</code>
+<code>  notify:endpoint &lt;https://websocket.example/subscription&gt; ;</code>
+<code>  notify:features notify:state, notify:rate, notify:expiration .</code>
+<code></code>
 <code>&lt;#webhookNotification&gt;</code>
-<code>    a notify:WebHookSubscription2021 ;</code>
-<code>    notify:endpoint &lt;https://webhook.example/subscription&gt; ;</code>
-<code>    notify:features notify:rate, notify:expiration, notify:webookAuth</code>
-<code>.</code>
-</pre>
+<code>  a notify:WebHookSubscription2021 ;</code>
+<code>  notify:endpoint &lt;https://webhook.example/subscription&gt; ;</code>
+<code>  notify:features notify:rate, notify:expiration, notify:webhookAuth .</code></pre>
 
                     <figcaption property="schema:name"><code>JSON-LD</code> serialization of a Solid Server Metadata Resource</figcaption>
                   </figure>
--- a/protocol.html
+++ b/protocol.html
@@ -297,10 +297,11 @@ content: "";
                 <li class="tocline">
                   <a class="tocxref" href="#introduction"><span class="secno">1</span> <span class="content">Introduction</span></a>
                   <ol>
-                    <li><a href="#use-cases"><span class="secno">1.1</span> <span class="content">Use Cases</span></a></li>
-                    <li><a href="#terminology"><span class="secno">1.2</span> <span class="content">Terminology</span></a></li>
-                    <li><a href="#overview"><span class="secno">1.3</span> <span class="content">Overview</span></a></li>
-                    <li><a href="#conformance"><span class="secno">1.4</span> <span class="content">Conformance</span></a></li>
+                    <li><a href="#overview"><span class="secno">1.1</span> <span class="content">Overview</span></a></li>
+                    <li><a href="#use-cases"><span class="secno">1.2</span> <span class="content">Use Cases</span></a></li>
+                    <li><a href="#terminology"><span class="secno">1.3</span> <span class="content">Terminology</span></a></li>
+                    <li><a href="#namespaces"><span class="secno">1.4</span> <span class="content">Namespaces</span></a></li>
+                    <li><a href="#conformance"><span class="secno">1.5</span> <span class="content">Conformance</span></a></li>
                   </ol>
                 </li>
                 <li class="tocline">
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
When is the RDF format to be decided?

For instance, rather than `notify:hasWebSocketNotification`, I wonder if we want `notify:hasNotificationChannel [ a notify:WebSocket ]` or similar.

acoburn:
The format is entirely open for discussion.

My initial thought was along the lines of `notify:hasNotificationChannel`, but I chose the protocol-specific properties purely from a performance consideration. If a client needs to follow links to many different remote resources, I worried that it would be too much of a burden on clients.

For instance, if the Storage Metadata resource contains the following:

```turtle
<>
    a solid:StorageMetadata ;
    notify:hasNotificationChannel <https://foo.example/Metadata>, <https://bar.example/Metadata>, <https://baz.example/Metadata> .
```

A client would need to dereference _all_ of the remote resources in order to determine which one is relevant. But the client may only be interested in WebSockets and those remote resources may be for WebHooks or LDN.

From an RDF data modeling perspective, I prefer the `notify:hasNotificationChannel` structure, but I don't really like that from an app development perspective.

RubenVerborgh:
> A client would need to dereference _all_ of the remote resources in order to determine which one is relevant.

True. To mitigate this, we should mandate certain information (such as typing) to be present in the shape.

acoburn:
> To mitigate this, we should mandate certain information (such as typing) to be present in the shape.

I like that suggestion a lot. I will incorporate it into this PR

acoburn:
@RubenVerborgh what do you think of the use of `rdf:value` in structured bnodes such as:

```turtle
<>
    a solid:StorageMetadata ;
    notify:hasNotificationChannel [
      a notify:WebSocketSubscription2021 ;
      rdf:value <https://foo.example/Metadata> ]
.
```

Or would you prefer an offset description like so:

```turtle
<>
    a solid:StorageMetadata ;
    notify:hasNotificationChannel <https://foo.example/Metadata>
.
<https://foo.example/Metadata>
    a notify:WebSocketSubscription2021 
.
```

RubenVerborgh:
I'm not a big fan to be honest, but I see the problem.
I wonder if we want something like `notify:accessUrl "https://foo.example/Metadata"` (the literal is deliberate). Would also enable authenticated URLs.

acoburn:
How about this: we start by keeping it simple.

IOW, all discovery metadata resides in the Storage metadata resource like this:

```turtle
<>
    a solid:StorageMetadata ;
    notify:hasNotificationChannel <#websocketNotification> 
.
<#websocketNotification>
    a notify:WebSocketSubscription2021 ;
    notify:endpoint <https://websocket.example/subscription> ;
    notify:features notify:state, notify:rate, notify:expiration 
.
```

This could be wrapped in a bnode as in your [example above](https://github.com/solid/notifications/pull/29#issuecomment-1004217794), if desired, but there would be no defined support for externally-defined metadata. So `<> notify:hasNotificationChannel <https://foo.example/Metadata>` would not be explicitly described at all.

RubenVerborgh:
That's great, blank node definitely not needed.
I'd just refer to #30 so we don't use "endpoint" 🙂

acoburn:
Thanks, @RubenVerborgh. FWIW, I cringed every time I used `endpoint` here, but I wanted to avoid combining this change with #30. But yes, I really want to remove the use of "endpoint".

protocol.html
```diff
@@ -486,81 +516,41 @@ <h3 property="schema:name">Discovery</h3>
...
-                    <figcaption property="schema:name"><code>POST</code> request including <code>type</code> and <code>features</code> targeting the Notifications Gateway.</figcaption>
-                  </figure>
-
-                  <figure id="notification-gateway-response" class="example listing" rel="schema:hasPart" resource="#notification-gateway-response">
-                    <p class="example-h"><span>Example</span>: Response from the Notification Gateway including a type of subscription endpoint with available features.</p>
-
-                    <pre about="#notification-gateway-response" property="schema:description" typeof="fabio:Script"><code>HTTP/2</code>
-<code>Content-Type: application/ld+json</code>
+<code>&lt;&gt;</code>
+<code>  a solid:StorageMetadata ;</code>
```
csarven, line 525: Just noting that something like this is what we'll probably end up with but there is a slight mismatch here with the description around it in that the resource is about server-metadata but `<>` is about storage metadata.

Same goes for the Protocol section starting off with "capabilities of a Solid Storage" but the text has emphasis on server.

protocol.html
```diff
@@ -383,6 +381,21 @@ <h2 about="#introduction" property="schema:name" typeof="deo:Introduction">Intro
                 <li><a href="http://data.europa.eu/esco/occupation/c40a2919-48a9-40ea-b506-1f34f693496d">Application developers</a> who want to implement a client to listen for updates to particular resources.</li>
               </ul>
 
+              <section id="overview" inlist="" rel="schema:hasPart" resource="#overview">
```
csarven, line 384: On one hand I like the Overview closer to the top of the section but it contains terminology that's not clear just yet. That was the initial reason to have it after Terminology.. and thought that the Use Cases section made sense before diving into the technical details..

(As for Use Cases, going forward, that can be truncated and have it rolled into Introduction.. and link to notifications-ucr document.)

csarven (approved):
I suppose we'll stick to solid-server-metadata-(resource, request, response) for the time being until https://github.com/solid/specification/issues/355 is sorted out. That might end up changing to storage-metadata-resource...

Also acknowledging issue #30 .. and endpoint is fine for now. It non-normative information as it stands any way. "uri" (or "target"?) seems fine.

kjetilk:
Allow me to advocate a change of perspective, which may or may not be relevant... 

There are a couple of long-lived abstractions, one is that we prefer self-describing resources, so we tend to prefer the URI to match the target URL, or differ only in a fragment identifier. The other is essentially an OOP-inspired object or tree that is usually anonymous, i.e. the subject isn't really important at all, all important information is in property and object position. But neither of these are absolutes, and in an RDF, the subject is actually important, the subject is what we're talking about.

It seems to me, and I admit to not have looked very carefully, that what we're talking about here is the URI that should be subscribed to. Making it RDFy, it should be the subject, e.g.

```turtle
<>
    a solid:StorageMetadata ;
    notify:hasNotificationChannel <https://websocket.example/subscription> 
.
<https://websocket.example/subscription>
    a notify:WebSocketSubscription2021 ;
    notify:features notify:state, notify:rate, notify:expiration 
.
```

In general, I'd like to see patterns like this more... I hope it fits. :-)

csarven:
I agree on describing https://websocket.example/subscription , but then for it to be fully self-describing, it should be an RDF document and respond to GET, HEAD. Exposing its description in another resource is not what's generally considered as self-describing.

kjetilk:
> I agree on describing https://websocket.example/subscription , but then for it to be fully self-describing, it should be an RDF document and respond to GET, HEAD. 

If it can do that, great.

> Exposing its description in another resource is not what's generally considered as self-describing.

and if it can't, I wouldn't take a dogmatic approach to self-description, and still do it even though it wouldn't be self-describing. It would be a better knowledge graph, and when there's tension between self-description and knowledge graphs, then I would rather side with the knowledge graph, because, if you don't, what use is RDF?

csarven:
I'm not aware of a strong reason as to why the subscription resource (endpoint) can't or shouldn't respond to GET, HEAD.

As per Solid Protocol #server-post-target-not-found :

>When a POST method request targets a resource without an existing representation, the server MUST respond with the 404 status code.

Moreover, the request semantics of POST targeting a non-container (subscription resource) is currently undefined. (We have considerations/WIP on RDF Merge... as you know.)

Hence, I propose that the Notification Protocol has a requirement for GET, HEAD requests targeting a subscription resource in that the server MUST respond with 200 or 405. I suggest 200. The metadata as described in this PR should be available from the subscription resource in addition to the storage metadata resource.

---

I suggest additional considerations mostly for the record but preferably followed up in a separate issue/PR, if need be, and definitely not block this PR!

* Use the storage resource instead of storage metadata resource to describe storage's notification channels. (Ultimately depends on issue 355.)

* (I love and hate this idea.) Change the subscription resource from a non-container to a container. It will work with existing behaviour on POST to container in Solid Protocol. This will open up the possibility for a server to track/record subscription requests (as notifications) in and of itself. The immediate concern with subscription resource as a container is that if the description is to be readable (200 to GET, HEAD as mentioned above), then the container would reveal containment statements - a "feature" we definitely don't want as is. To get around this, essentially there needs to be a requirement that would, if at all, only allow the owner of the storage to have access to these subscription requests. Alternatively, a requirement that would allow servers to omit containment statements. On the other hand, if GET, HEAD to subscription resource is a 405, we don't have that problem.

elf-pavlik (approved):

PR #31, opened 2022-01-13
Define StreamingHTTPSubscription2021 type for Solid Notifications

elf-pavlik:
This document defines the requirements for `StreamingHTTPSubscription2021.

### DIFF
--- a/streaming-http-subscription-2021.md
+++ b/streaming-http-subscription-2021.md
@@ -121,9 +121,9 @@ And define how to handle notifications
 
 ## Authentication and Authorization
 
-Streaming HTTP Subscription has advantage of being able to authenticate with notification endpoint
-not only subscription endpoint. Same *access token* can be used with both endpoints.
-This doesn't just rely on notification endpoint being a [Capability URL](https://www.w3.org/TR/capability-urls/)
+Streaming HTTP Subscription has the advantage of being able to authenticate with the notification endpoint
+not only the subscription endpoint. The same *access token* can be used with both endpoints.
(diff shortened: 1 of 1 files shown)


streaming-http-subscription-2021.md
```diff
@@ -0,0 +1,134 @@
...
+            console.log(decoder.decode(value))
+            return reader.read().then(handle)
+        })
+      })()
+
+```
+
+## Authentication and Authorization
+
+Streaming HTTP Subscription has advantage of being able to authenticate with notification endpoint
```
acoburn, line 124:
```suggestion
Streaming HTTP Subscription has the advantage of being able to authenticate with the notification endpoint
```

streaming-http-subscription-2021.md
```diff
@@ -0,0 +1,134 @@
...
+not only subscription endpoint. Same *access token* can be used with both endpoints.
```
acoburn, line 125:
```suggestion
not only the subscription endpoint. The same *access token* can be used with both endpoints.
```

streaming-http-subscription-2021.md
```diff
@@ -0,0 +1,134 @@
...
+This doesn't just rely on notification endpoint being a [Capability URL](https://www.w3.org/TR/capability-urls/)
```
acoburn, line 126:
```suggestion
This doesn't just rely on the notification endpoint being a [Capability URL](https://www.w3.org/TR/capability-urls/)
```

acoburn (approved):
LGTM - some minor inline textual corrections

Issue #30, opened 2022-01-03
Replace 'endpoint' with API


acoburn:
For an ecosystem that is based on capability/feature discovery, it is arguably better to think less about "endpoints" and more about "hypermedia APIs". As such, I would propose removing instances of 'endpoint' in this specification and replacing them with either 'API' or 'URI', as appropriate.

TallTed:
"Endpoint" and "API" are not interchangeable terms.  Neither are "API" and "URI", nor "endpoint" and "URI".

"Endpoint" is well understood as the "agent to which an API call (including an HTTP method call, for which a resource URI is typically an input) is issued".

An "API" provides function signatures, including inputs and outputs.  These might be passed within a URI (typically as part of the query segment) for a simple HTTP `GET`, or in the payload of an HTTP `POST`.

There is nothing that says that a Solid server cannot respond to an incomplete (including empty) API call with the signature of that API, on which the original caller may act by including any missing inputs in a subsequent call.

[OpenAPI 3.0 (née Swagger)](https://swagger.io/specification/) seems well suited to the requirements here, and I'm rather surprised that it's not being put to use.

elf-pavlik:
I don't think we should rely on nuances of endpoint vs API. Personally, I think we should make a distinction between

* subscription URL - which always requires AuthN and returns subscription details
* notification URL - in most cases a capability URL tied to AuthN from subscription
  * notification source - current subscription type call it 'endpoint' - a subscriber can establish a connection with the publisher to receive notifications
  * notification target - currently only WebHooks subtype defines it, it doesn't use source since notifications are being sent by publisher establishing a connection with subscriber

elf-pavlik:
To clarify
* **source** would be used instead of *endpoint* in:
  * WebSockets Subscription type - `wss:`
  * EventSource Subscription type - `https:`
  * StreamingHTTP Subscription type - `https:`
 * **target** would be used in:
   * WebHook Subscription type 

Once @csarven proposes LDN Subscription type we would see which one applies there, I would guess it would provide some inbox as *target* so probably I also wouldn't use *source*. Just guessing here without seeing the actual LDN sub type proposal.

elf-pavlik:
#43 is addressing this issue

PR #33, opened 2022-01-22
Source and Target flows

elf-pavlik:
As discussed in #30 this PR starts describing both flows. Naming is still under discussion and it very likely will change.

Generated sequence diagrams:

**source flow**

Screen Shot 2022-01-22 at 16 11 42

**target flow**

Screen Shot 2022-01-22 at 16 12 04


csarven:
You have it backwards. No one is required to use a random diagram tool or learn a specific tool's language to generate diagrams or carry its source around in a repository or expect that it should be tied GitHub actions or require Javascript or whatever... It is not only a distraction but unnecessary overhead. And once again, it is also not at all aligned with the shift towards having the specifications readable and writeable from a Solid server using Solid applications ( https://github.com/solid/specification/blob/main/meetings/2022-01-05.md#solid-protocol-publication-and-access , https://github.com/solid/team/blob/main/meetings/2022-01-12.md#solid-protocol-publication-and-access ,.. ). You're welcome to use whatever tool suits you to generate the SVG (or a suitable/common image format). SVG that can be linked from the specification is the baseline / common / accessible language.

csarven:
"subscription container" was originally mentioned in https://github.com/solid/notifications/pull/29#issuecomment-1009840586 . That possibility can be hashed out in https://github.com/solid/notifications/issues/36

acoburn (approved):

acoburn:
The plan is, in a separate PR, to add the GH-Action flow to auto-generate the SVG resources from these mermaid documents.

csarven:
I don't understand why this got merged especially the prior state was a DRAFT. It went from a comment in an issue to "as mentioned in" as if there is some wide consensus.

@elf-pavlik , is this "flow" backed by an implementation experience that you have? Can you provide links?

Also.. can we not have filenames including "mmd" in the name (`source-flow.mmd.svg`)?

elf-pavlik:
> Also.. can we not have filenames including "mmd" in the name (source-flow.mmd.svg)?

In the following PR I will add GitHub workflow to auto-generate svg from mmd. So there will only be `*.mmd` files in the repo, `mmd.svg` will be only generated in gh-pages and used in URLs.

Do you see any problems with having `*.mmd.svg` in opaque URLs ?

acoburn:
Can we please not make this conversation about the specific name of the SVG files? If `*.mmd.svg` is truly a problem (I do not think it is), that can be changed separately, but that is orthogonal to the issue here of providing diagrams outlining these two high-level flows.

csarven:
I assumed we didn't need to discuss "what to leave out" ( https://www.w3.org/Provider/Style/URI ) so I simply asked for "mmd" to be removed because it seemed like an obvious thing.

That particular format/tooling may not at all be used to generate the SVGs in the future and I see absolutely no reason to carry that baggage around in the specifications.

If you don't want to "leave it out" because it is inconvenient for you, then say that up front as your reason. I suppose if you blasted your name all over in the filename you'd still argue that it is just "opaque URL"?

Please remove "mmd" or any software-specific name that happens to slip in. We don't need to discuss further.

---

@elf-pavlik , is this flow backed by an implementation experience that you have? Can you provide links?

PR #41, opened 2022-03-10
Remove incorrect table of contents item

acoburn:
The use cases section was removed earlier, but the ToC still refers to it.


elf-pavlik (approved):

csarven (approved):
Editors don't need to PR "minor" changes along these lines.

Issue #44, opened 2022-03-17
Documents including script from dokie.li fail to load.


elf-pavlik:
I think we should serve that `.js` form github so it is more reliable.

I had the same issue a while ago and mentioned it during one of the earlier meetings.

![Screen Shot 2022-03-16 at 18 57 33](https://user-images.githubusercontent.com/876431/158715408-b47903e9-e964-45b4-8b9d-b55403b0c0dd.png)

@csarven 

related: https://github.com/solid/notifications-panel/issues/36

elf-pavlik:
I don't think my browser is blocking anything. That script from `dokie.li` usually loads just fine but at times it just fails to load. It just looks like `dokie.li` doesn't work in a reliable manner. Do you have some kind of CDN in front of it?

elf-pavlik:
@csarven please take a look at #50 

> I'd like to understand your issue better. I'd be grateful if you know how we can reduce the size: https://github.com/linkeddata/dokieli

I would suggest using https://chrisbateman.github.io/webpack-visualizer/ to see what gets bundled into that file.

It looks like docile bundles a ton of things. Which of them do you actually use for the spec?

csarven:
I look forward to the day where I can help you understand dokieli and the notion/point of dogfooding. But, today may not be that day.

Thanks for the feedback on the webpack visualizer. We've used such tools back in 2016.. but the situation is a bit more complicated than what appears on the surface - you'll have to see the dirty details... 

Once again, you haven't shown how we can reproduce your error. And I'm genuinely trying to resolve it.

elf-pavlik:
> Can you share the screenshot of your Network tab?
> Are you experiencing the same error on other Web browsers?

This error isn't persistent, it just happens now and then. After #50 I will not even be noticing anymore when that script fails to load.

> I look forward to the day when I can help you understand dokieli and the notion/point of dogfooding. But, today may not be that day.

Please do help me understand, what does I or anyone else who **just wants to read the spec** need that dokieli script for?

csarven:
Closing this issue as 50 seems to help. You can also consider block the script or disable JS.

PR #50, opened 2022-03-17
make dokieli.js async

elf-pavlik:
A partial solution to #44 

To see the difference:
1. Open a new browser tab
2. Open dev tools
  * set throttling to 2G
  * disable cache
3. Paste https://solid.github.io/notifications/protocol in the address bar and navigate

The page will not render until the whole `dokieli.js` gets loaded, which with throttling set to 2G will take a while.

For the version in this PR, repeat steps 1. & 2.
3. Paste https://elf-pavlik.github.io/notifications/protocol in the address bar and navigate

The page will render before `dokieli.js` finishes loading.

To fully solve #44 we should follow Ruben's suggestion captured in https://github.com/solid/notifications-panel/issues/36 and inject stylesheet from the javascript.

### DIFF
--- a/eventsource-subscription-2021.html
+++ b/eventsource-subscription-2021.html
@@ -125,7 +125,7 @@ content: "";
 }
 </style>
     <link href="https://dokie.li/media/css/dokieli.css" media="all" rel="stylesheet" />
-    <script src="https://dokie.li/scripts/dokieli.js"></script>
+    <script src="https://dokie.li/scripts/dokieli.js" async></script>
   </head>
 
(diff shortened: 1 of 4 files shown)


csarven (approved):
Great, thanks!

But I don't quite understand how this PR partially addresses the issue you brought up about script failing to load - connection dropped / timeout? so the whole script wasn't fetched? I'll have to dig into this (unless you have links handy) but I'm not sure if script failing to load blocks everything else - that seems draconian. Wouldn't it just continue parsing and render? Again, the JS in not required for the content to be loaded . JS can even be disabled.

elf-pavlik:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#notes

> Scripts without [`async`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-async) , [`defer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-defer) or `type="module"` attributes, as well as inline scripts, are fetched and executed immediately, before the browser continues to parse the page.

csarven:
Can you share the screenshot of your Network tab?

Are you experiencing the same error on other Web browsers?

PR #54, opened 2022-03-24
Clarify discovery mechanism as independent of Solid Metadata Resources

acoburn:
This removes the concept of a "Solid Storage Metadata Resource" (which doesn't exist). Should such a concept be introduced in the Solid Protocol, any relationship to notifications can be added there.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -310,7 +310,7 @@ content: "";
                       <a class="tocxref" href="#discovery"><span class="secno">2.1</span> Discovery</a>
                     </li>
                     <li class="tocline">
-                      <a class="tocxref" href="#notification-subscription"><span class="secno">2.2</span> Notification Subscription</a>
+                      <a class="tocxref" href="#subscription"><span class="secno">2.2</span> Subscription</a>
                     </li>
                   </ol>
                 </li>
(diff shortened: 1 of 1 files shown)


csarven (approved):
Looks good.

Unless I've overlooked something, where to actually obtain the "Notification Subscription Metadata Resource" is not explained. As it stands, it is presumably a resource different than the subscription resource and the "topic" resource. Although the subscription metadata could be exposed anywhere, we need describe the one that's required - possibly associated with the "topic" resource and have it be discoverable via a link relation. `describedby` may be one simple option as the relation in the HTTP Link header.

We can come back to this after this PR.

acoburn:
> Unless I've overlooked something, where to actually obtain the "Notification Subscription Metadata Resource" is not explained

That's correct. The way I've been thinking about this is that a Solid resource would point to this metadata resource, either via a link header on an individual resource or via the (possible future) Storage metadata resource. I was tempted to articulate that as part of this PR but then I didn't want this to depend on a mechanism that isn't yet defined in the Solid Protocol. We can definitely come back to this point because it is a notable gap.

The other issue is whether the metadata resource is separate from the subscription resource, and if so, whether that's discoverable via link header. I like the suggestion of `describedby`, but we also should sort out whether the metadata resource is, in fact, separate from the subscription resource. (I'm open to ideas)

elf-pavlik (approved):

PR #56, opened 2022-03-25
added Turtle with anotations for specification-tests

elf-pavlik:
It adds annotations following the  Turtle example in: https://github.com/solid/specification-tests#annotations
It follows embedding in HTML as recommended in Turtle spec: https://www.w3.org/TR/turtle/#in-html


elf-pavlik:
I'm going to merge this PR later today

PR #59, opened 2022-04-04
Guidance to Subscription Type Publishers 1

CxRes:
We add text guiding Subscription Type publishers to register the Subscription Type information with the Solid project.

This is the first part of the action item on author guidance from the notifications-panel meeting on 31-03-2022, only containing subscription type registration guidance. Guidance on how one might go about creating a new protocol shall be added later (might require us to write a primer first before linking it from the standard).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,15 @@ These reports are incubated by the [Solid Notifications Panel](https://github.co
 The Solid Notification Protocol makes it possible to define any number of subscription types.
 The list of subscription types below does not intend to be comprehensive. This list is intended
 to be useful to application developers looking for existing subscription types as well as to
-authors of new subscription types. New subscription types may be added here, but there is no
-requirement to do so.
+authors of new subscription types.
 
 * [WebSocketSubscription2021](https://solid.github.io/notifications/websocket-subscription-2021) Editor's Draft.
 * [StreamingHTTPSubscription2021](https://solid.github.io/notifications/streaming-http-subscription-2021) Editor's Draft.
 * [EventSourceSubscription2021](https://solid.github.io/notifications/eventsource-subscription-2021) Editor's Draft.
 * [LinkedDataNotificationsSubscription2021](https://solid.github.io/notifications/linkeddatanotifications-subscription-2021) Editor's Draft.
 * [WebHookSubscription2021](https://github.com/solid/notifications/blob/main/webhook-subscription.md) Editor's Draft.
+
+New subscription types may be added here, but there is no requirement to do so. Adding a subscription type to the list above aids in its discoverability by the solid community. We request authors of new subscription types to open a PR with this repository, so that the subscription type they have authored can be included in the list above.
 
 ## Diagrams
 
--- a/protocol.html
+++ b/protocol.html
@@ -805,6 +805,12 @@ content: "";
 
                     <figcaption property="schema:name">Subscription request with a custom subscription type.</figcaption>
                   </figure>
+
+                  <p>
+                    While not necessary, we encourage authors to register subscription types with the Solid project, allowing them to be discovered by the Solid community.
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -12,15 +12,16 @@ These reports are incubated by the [Solid Notifications Panel](https://github.co
...
-requirement to do so.
+authors of new subscription types.
 
 * [WebSocketSubscription2021](https://solid.github.io/notifications/websocket-subscription-2021) Editor's Draft.
 * [StreamingHTTPSubscription2021](https://solid.github.io/notifications/streaming-http-subscription-2021) Editor's Draft.
 * [EventSourceSubscription2021](https://solid.github.io/notifications/eventsource-subscription-2021) Editor's Draft.
 * [LinkedDataNotificationsSubscription2021](https://solid.github.io/notifications/linkeddatanotifications-subscription-2021) Editor's Draft.
 * [WebHookSubscription2021](https://github.com/solid/notifications/blob/main/webhook-subscription.md) Editor's Draft.
 
+New subscription types may be added here, but there is no requirement to do so. Adding a subscription type to the list above aids in its discoverability by the solid community. We request authors of new subscription types to open a PR with this repository, so that the subscription type they have authored can be included in the list above.
```
csarven, line 23: There is a bit of redundancy (and also with the earlier paragraph). What do you think of this:

```suggestion
To add a new subscription type to the list, please create a PR including the name and the URL of the document describing the subscription type. The Solid CG reserves the right to include or update the list.
```

protocol.html
```diff
@@ -805,6 +805,12 @@ <h3 property="schema:name">Subscription Types Extension</h3>
 
                     <figcaption property="schema:name">Subscription request with a custom subscription type.</figcaption>
                   </figure>
+
+                  <p>
+                    While not necessary, we encourage authors to register subscription types with the Solid project, allowing them to be discovered by the Solid community.
+                    The process for registration shall be avaiable at the <a href="https://github.com/solid/notifications">Solid Notifications Protocol Repository</a>
```
csarven, line 811: I don't think we should introduce terminology like register/registration - it sounds more heavy duty process than what's intended I think. Would you consider rephrasing this? What you had in the README and also what I suggested is closer to what we want.

csarven, line 811:
```suggestion
                    The process for registration shall be available at the <a href="https://github.com/solid/notifications">Solid Notifications repository</a>.
```

csarven (changes requested):

README.md
CxRes: I like your language better!

protocol.html
CxRes: Does "inform" sound better? I can draft some language accordingly!

README.md
CxRes: Now that I think about it, would we like to have a leading line saying something like,"We welcome and encourage community contributions of subscription types."

csarven: I think that's implied by "to add a new subscription.."

protocol.html
csarven: I think the text need not be too different between the README and what's in the spec. We can potentially copy/paste.

elf-pavlik: copy/paste sounds ok

csarven:
Approved on the condition to follow-up on action in https://github.com/solid/notifications-panel/blob/main/meetings/2022-04-28.md#guidance-to-subscription-type-publishers-1

csarven:
Follow up: https://github.com/solid/notifications/commit/d19b943fbf2f4fcf66088fac9af1c5f0a5e7ccf6 . There may be changes depending on discussions/review involving index (or registry) for subscription types as well as the resolution of https://github.com/solid/notifications/issues/66 .

CxRes:
@csarven Please look at the use of "register" (protocol.html/L817) which everyone was so unhappy about. I had proposed some language in the PR comments and gitter.

PR #60, opened 2022-04-06
Define Notification Content

jaxoncreed:
I don't know if there were already discussions around this, but I thought that the notifications content needed to be defined more than just referring to Activity Streams. But, I may be wrong.

### DIFF
--- a/notification-content-2022.md
+++ b/notification-content-2022.md
@@ -1,0 +1,156 @@
+# Notification Content 2022
+
+## 1. Abstract
+The notificaions protocol requires that:
+
+> The content of a Solid Notification is defined in terms of a Activity Streams 2.0 [ACTIVITYSTREAMS-CORE]. As such, JSON-LD 1.0 is a required baseline serialization format.
+
+However, this is still a broad perscription that causes implementation details to be up to interpretation. Subscribing clients want to be sure that the content of a notification is the same for each action. Notification-Content-2022 defines the exact content that should be expected from each action on a Solid Pod.
+
+## 2. Status of the Document
+
+TODO
+
+## 3. Notification Feature
+
+> This is proposed as a feature in this spec, but it could also be easily inserted into the main protocol.
+
+Notification Content 2022 is a notification feature. As such, any compliant storage server MUST list it among other features during the notification discover phase as the URL `http://www.w3.org/ns/solid/notifications#NotificationContent2022`.
+
+For example, a compliant storage server could have a document like the following:
+
+```turtle
+@prefix notify: <http://www.w3.org/ns/solid/notifications#> .
+
+<>
+  notify:notificationChannel <#websocketNotification> .
+
+<#websocketNotification>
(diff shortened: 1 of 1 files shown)


elf-pavlik (approved):
I think this is a good first ED to start iterating over.

csarven (changes requested):
* Notifications Protocol includes a core section for the Data Model. I suggest updating that section instead of a separate work item. If/when the Data Model becomes more prominent/complicated, we can separate.

* Notifications are not limited to RDF resources. It can be for any type.

* We shouldn't use LDP vocabulary - it has a specific/constrained view on RDF sources that's not shared in the Solid Protocol.

* notify:NotificationContent2022 as a "feature" seems a bit strange to me.

* Again, I'd very much like us to align/reuse AS's "object" (as you're in this document for the notification)... the current spec uses WebSub's "topic". We can take this up separately: https://github.com/solid/notifications/issues/62

notification-content-2022.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+## 4. Notifications for Events
+Each time an event happens on a storage server, one or more notifications will be sent to the subscribing client in the following formats.
+
+A notification activity MUST include a [`type`](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-type) field mapped to an [activity type](https://www.w3.org/TR/activitystreams-vocabulary/#types).
+
+A notification activity SHOULD include an [`id`](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-id) field.
+
+If the subscription was initiated by an authenticated WebId, a notification activity MUST include an [`actor`](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-actor) field mapped to the the WebId that initiated the subscription.
```
jaxoncreed, line 41: Say that it's a uri that represents the actor

protocol.html
```diff
@@ -723,38 +723,124 @@ <h2 property="schema:name">Notification Data Model</h2>
...
+                </div>
+              </div>
+
+              <p>Each time an event happens on a storage server, one or more notifications will be sent to the subscribing client in the following formats.</p>
+
+              <p>A notification activity MUST include a <a href="https://www.w3.org/TR/activitystreams-vocabulary/#dfn-type">type</a> field.</p>
+
+              <p>A notification activity SHOULD include an <a href="https://www.w3.org/TR/activitystreams-vocabulary/#dfn-id">id</a> field.</p>
+
+              <p>A notification activity MUST include an <a href="https://www.w3.org/TR/activitystreams-vocabulary/#dfn-actor">actor</a> field mapped to the URI of the authenticated agent that created the subscription. If an authenticated agent did not create the subscription, the actor field is not required.</p>
```
acoburn, line 739: There are privacy considerations here. _Requiring_ this field is, in my opinion, too strong. I would prefer to see this at a `MAY` or `SHOULD`, but definitely not `MUST`.

acoburn (changes requested):

PR #61, opened 2022-04-06
Update Delta Feature

jaxoncreed:
This is a proposal for including the changes made (delta) in the notification.

### DIFF
--- a/update-delta.md
+++ b/update-delta.md
@@ -1,0 +1,129 @@
+# Update Delta
+
+## 1. Abstract
+When an RDF resource is updated on a storage server, a notification is sent to a subscribing client, but this notification does not tell the client how the resource was updated. As a result, the client must make another request to the storage server to retieve the document and reconcile its updates. Update Delta is a notifications feature that tells the client change, or delta between the document pre-update and post-update in the notification activity.
+
+## 2. Status of the Document
+TODO
+
+## 3. Notification Feature
+Update Delta is a notification feature. As such, any compliant storage server MUST list it among other features during the notification discover phase as the URL `http://www.w3.org/ns/solid/notifications#updateDelta`.
+
+For example, a compliant storage server could have a document like the following:
+
+```turtle
+@prefix notify: <http://www.w3.org/ns/solid/notifications#> .
+<>
+  notify:notificationChannel <#websocketNotification> .
+<#websocketNotification>
+  a notify:WebSocketSubscription2021 ;
+  notify:subscription <https://websocket.example/subscription> ;
+  notify:features notify:state, notify:rate, notify:expiration, notify:updateDelta .
+```
+
+## 4. Update Activity
+
+Any Update activity (including updates to containers) MUST include the [`result`](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-result) field.
+
+> Is the `result` field the most appropriate one for this? I didn't see an ActivityStreams property that was really appropriate. Should we make a new property?
+
+The `result` field MUST be mapped to [`Add` type](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-add) and/or [`Remove` type](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-remove) activities.
+
+The `object` field of the `Add` and `Remove` activities MUST map to a graph distinct from the default graph.
+
+Each graph referenced by the `Add` and `Remove` activities must contain the respective triples that were added and removed.
+
+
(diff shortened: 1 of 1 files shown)


elf-pavlik (approved):
I think it serves as an interesting direction to explore.

I think we should still consider another option in parallel, for example, reusing [Modifying Resources Using N3 Patches](https://solidproject.org/TR/protocol#n3-patch)

As always with RDF and deltas, Blank Nodes will start biting sooner or later. We may want to recommend that storage which uses deltas should [skolemize](https://www.w3.org/TR/rdf11-concepts/#section-skolemization) all the blank nodes in stored resources. Yet another reason to coordinate any work related to deltas and patches. Regarding skolemizing blank nodes we could ask for feedback from https://lodlaundromat.org/

jaxoncreed:
Look into RDF Delta

elf-pavlik:
@joachimvh do you have some suggestions based on how you see implementing it in CSS? We only look at RDFSources here and including the delta in the notification message.

As I mention in my previous comment diffing RDF always comes with the nuance of labeling blank nodes and preferably we are reusing existing strategy to deal with it.

I think if the resource was updated with HTTP PATCH the same patch could be considered a delta, for updates with HTTP PUT the storage server would need to calculate the diff.

joachimvh:
> @joachimvh do you have some suggestions based on how you see implementing it in CSS?

Skolemizing blank nodes seems drastic as you are modifying the user's data. If someone didn't want their deltas to include blank nodes it's up to them to not put blank nodes in their data. Or a server could reject RDF data containing blank nodes if you really wanted to prevent them.

Actually calculating and emitting deltas in CSS would probably not be trivial. In the case of a PUT the data gets replaced directly without the server looking at the actual data. In the case of a PATCH this is internally done by doing a read followed by a write, and the same event gets emitted as in the case of a PUT.

update-delta.md
```diff
@@ -0,0 +1,129 @@
+# Update Delta
+
+## 1. Abstract
+When an RDF resource is updated on a storage server, a notification is sent to a subscribing client, but this notification does not tell the client how the resource was updated. As a result, the client must make another request to the storage server to retieve the document and reconcile its updates. Update Delta is a notifications feature that tells the client change, or delta between the document pre-update and post-update in the notification activity.
+
+## 2. Status of the Document
+TODO
```
elf-pavlik, line 7:
```suggestion
**Version: 0.1**
```

elf-pavlik (approved):
I think we shouldn't have PR left hanging. We can mark it as 0.1 and merge it. This should signal that this feature could possibly completely change. At least we will have a draft to iterate on.

update-delta.md
```diff
@@ -0,0 +1,129 @@
+# Update Delta
+
+## 1. Abstract
+When an RDF resource is updated on a storage server, a notification is sent to a subscribing client, but this notification does not tell the client how the resource was updated. As a result, the client must make another request to the storage server to retieve the document and reconcile its updates. Update Delta is a notifications feature that tells the client change, or delta between the document pre-update and post-update in the notification activity.
```
TallTed, line 4:
```suggestion
When an RDF resource is updated on a storage server, a notification is sent to a subscribing client, but this notification does not tell the client how the resource was updated. As a result, the client must make another request to the storage server to retieve the document and reconcile its updates. Update Delta is a notifications feature that tells the client the change, or delta, between the document pre-update and post-update in the notification activity.
```

update-delta.md
```diff
@@ -0,0 +1,129 @@
...
+
+## 2. Status of the Document
+TODO
+
+## 3. Notification Feature
+Update Delta is a notification feature. As such, any compliant storage server MUST list it among other features during the notification discover phase as the URL `http://www.w3.org/ns/solid/notifications#updateDelta`.
```
TallTed, line 10:
```suggestion
Update Delta is a notification feature. As such, any compliant storage server MUST list it among other features during the notification discovery phase as the URL `http://www.w3.org/ns/solid/notifications#updateDelta`.
```

TallTed (changes requested):
Minor fixes

elf-pavlik:
During a recent call, we discussed it shortly again and were converging in a direction of using the same patch format which is being used for patching resources.
I think we might encounter some challenges when it comes to content type. While the patch might be using `text/n3` or `application/sparql-update` I don't see how it would be included in a notification using `application/ld+json` or `text/turtle`. Possibly this feature might be only available for `text/n3 with N3 Patch.

Issue #63, opened 2022-04-07
Rename `subscription resource`


uvdsl:
Although it seems to me that `subscription resource` is a long used and pretty set in stone term,
I would like point out that the naming may be confusing:

From the term alone, one could get the idea that a `subscription resource` is a resource whose representation describes a subscription. Especially, when modelling the resource where requests are submitted to as a container (#36), or even more, when considering materializing requests for subscriptions in such a container.

In my mind, what is currently referred to as a `subscription resource` is a resource whose representation describes a service that provides the subscription/notification functionalities. (independent of that resource being also a container or can be POSTed to)

Thus, I would like to propose the terms `subscription service resource` which is described by its `subscription service representation`.

csarven:
The notion of container is not part of Solid Notifications Protocol. "Resources" all the way down.

The subscription request (POST) and the response representations use the notification-channel-data-model and relevant notification-channel-types data model.

I don't have a strong opinion on re-naming, i.e., whether to use subscription resource or subscription service or subscription service resource. I would just say that the spec currently uses:
* topic resource
* description resource
* subscription resource
* notification channel

I find "resource" to be sufficient/simple. Notification channel may need to be referred as the notification channel resource, where applicable. Or, as you suggest, change subscription resource to subscription service resource.

Do others have thoughts/suggestions on this? We can still change / have more consistency.

uvdsl:
Just to clearify, I do not take issue in the usage of the term resource but the ambiguity of the term `subscription resource` as it may be confused with a resource representing a subscription.

As the current notion of `subscription resource` refers to a resource where one can create/register a subscription via HTTP POST, I feel we could be more precise in terms of naming here.

CxRes:
I would second @uvdsl here. Furthermore, if we were to go down the container (containing resources for a subscription) route in the future, we would have the term "subscription resource" available to us.

CxRes:
My preference is "Subscription Service", which might then also be referred to as "Subscription Service Resource" where we need to make it explicit that we are referring to an HTTP resource.

csarven:
I'd welcome renaming to "subscription service", but I propose an alternative view that can potentially tie up some loose ends:

The behaviour of subscription resource/service resembles a specialisation of LDN inbox with a particular data model. The inbox accepts subscription requests with optional features about a topic.

The request can be further specified with particular activity types (or shapes, see `activityType` proposal in https://github.com/solid/notifications/pull/147 ).

[...]

Issue #62, opened 2022-04-07
:bike: shed: Use `object` (closely aligned with AS) instead of `topic` to receive notifications about a resource


csarven:

csarven:
As of this writing:

`topic` is used in #subscription-request payload.
`object` is used in #notification-data-model (because of AS2).

I suspect that the core notification data model will need to express the topic/object of the notification. It may help to simplify the relationship between the terms used in the subscription request and the notification by using one. If the notification data model doesn't use AS2, something at least partly equivalent needs to be expressed about the topic/object of the notification.

csarven:
The semantics of AS2's `object` is not particularly accurate in the case of a subscription request - unless we introduce SubscriptionRequest as a type of as:Activity.

If notification context's JSON-LD defines `object` to be different than AS2's `object`, then it will throw an error due to protection rules ( https://www.w3.org/TR/json-ld11/#protected-term-definitions ) in notification context. It may be possible to define `object` such that it is not projected but, why bother...

The simpler solution may be to continue using topic and object separately for different purposes.

We can however clarify the relationship between topic and object in that they correspond to the same resource. PR can follow.

Issue #36, opened 2022-01-27
Subscription resource as a container


csarven:
Consider whether it is useful to define subscription resource as a container. Originally mentioned in https://github.com/solid/notifications/pull/29#issuecomment-1009840586

uvdsl:
I would like to provide the following take (which I used in my [WebPush Demo](https://github.com/uvdsl/solid-web-push#unsubscription)):
The `subscription resource` is a resource describing the subscription service(s) offered. It is your typical RDF resource. Its RDF representation links to the `subscription container` (aka. inbox of the subscription service), which is an LDP container.
With that, we keep the semantics we want for the description of the resource (`GET`,`HEAD`,`OPTIONS`) as well as a specified semantic (by the LDP) of the `POST` on the `subscription container`.

elf-pavlik:
I think we should keep in mind that based on the Subscription **Request**, the server will create Subscription **Response**. Currently, neither request nor response is materialized as a resource. I think we should always keep in mind that both of them play slightly different roles. The server has a big degree of freedom in how based on the request it will create the response, each subscription type can define certain requirements for both.

I think if we talk about representing Subscription Request as a resource, modeling it as resources in ldp:Container which could be used as LDN Inbox sound reasonable. Except for the expected response

> https://www.w3.org/TR/ldp/#h-ldpc-http_post

> 5.2.3.1 LDP clients SHOULD create member resources by submitting a representation as the entity body of the HTTP POST to a known LDPC. If the resource was created successfully, [LDP servers](https://www.w3.org/TR/ldp/#dfn-ldp-server) MUST respond with status code 201 (Created) and the Location header set to the new resource’s URL. Clients shall not expect any representation in the response entity body on a 201 (Created) response. 

Since Subscription Response is a crucial part of Solid Notifications, it may be a stretch trying to model requesting a subscription as LDP Container.

------

Since the Subscription Response represents the actual subscription created by the server based on the Subscription Request. We should also consider if we want to represent those as resources. To handle unsubscribing with DELETE they seem more appropriate than resources that would represent Subscription Request.

uvdsl:
_re Subscription Response:_ Fair point! If sub request and sub response must never materialise as a resource (as the POST response would include a body), then a non-LDP API may be better.

However, in case one would want to reuse LDP also for the notification protocol, one might want to materialise the subscription (itself) as a resource (or vice versa). This provides even an option for an asynchronous processing of subscriptions.
[...]

Issue #35, opened 2015-10-07
Live Updates and Push


bblfish:
The WebApp WG is working on a [Push API](http://www.w3.org/TR/push-api/). Since SoLiD has a section on [Live Updates](https://github.com/solid/solid-spec#live-updates) it would be useful to understand how these two complement each other, work together, what ideas can be taken from it, etc...

Some resources:
- the blog post [Push Notifications on the Open Web](http://updates.html5rocks.com/2015/03/push-notificatons-on-the-open-web)
- the IETF draft [Generic Event Delivery Using HTTP Push](https://datatracker.ietf.org/doc/draft-thomson-webpush-protocol/?include_text=1)

Any thoughts?

elf-pavlik:
I was thinking about it a little bit a few years ago: https://forum.solidproject.org/t/push-notifications-push-api/1079

I still think that this would require the application (OAuth client) to have a server-side component. This server-side component could use one of the *target* based #33 subscription types to subscribe to storage.
From there it would be pretty much that application implementation detail how Push API is being used.

Especially given Push API being a big part of what drives UX of the app, I think how it's being used should be fully up to the application developer.

In other words, Push API would be an implementation detail of each application using it. I don't think we need to specify anything directly related to it. What we need:

* *target* based subscription types which server-side component of the application would use.
* run use case of application (OAuth clients) which have both server-side and on-device components through AuthN and AuthZ panels.  I think the End-user shouldn't be burdened with the detail of certain clients having two components so bother components should use the same Client ID and be considered the same application.

uvdsl:
Hi all,

I built a demo for Web Push Notifications from Solid Pods.

The demo consists of three components:
- a Solid Pod (Server)
- which is extended with a [Web Push module](https://github.com/uvdsl/solid-web-push/)
- such that my [Progressive Web App](https://github.com/uvdsl/solid-web-pwa/) can have Web Push Notifications

In addition to the readme at the [Web Push module](https://github.com/uvdsl/solid-web-push/),
we have a supplementary [demo website](https://uvdsl.solid.aifb.kit.edu/conf/2022/icwe/demo) with a demo walkthrough, a sequence diagram, the data model and a [demo video](https://uvdsl.solid.aifb.kit.edu/conf/2022/icwe/demo#v).

How it works in a nutshell:
- The Web Push module has an inbox (at the Pod) where subscriptions can be posted to by a PWA.
- These subscriptions contain the target resource, the PWA would like receive Push Notifications on.
[...]

PR #64, opened 2022-04-15
Define WebPushSubscription2022 type for Solid Notifications

uvdsl:
Dear all,

I propose a first draft for defining the `WebPushSubscription2022` type, based on my [original submission](https://uvdsl.solid.aifb.kit.edu/conf/2022/icwe/demo), as decided in the Notifications Panel meeting on 2022-03-22 (see https://github.com/solid/notifications-panel/pull/52).

This PR addresses #35.

While writing this draft, I noticed that several existing issues need to be resolved prior to finishing this draft, including but not limited to  #36, #62 and #63.

I  deeply appreciate any feedback.

### DIFF
--- a/webpush-subscription-2022.bs
+++ b/webpush-subscription-2022.bs
@@ -1,0 +1,360 @@
+<pre class='metadata'>
+Title: Solid WebPushSubscription2022
+Boilerplate: issues-index no
+Local Boilerplate: logo yes
+Shortname: solid-webpush-subscription-2022
+Level: 1
+Status: w3c/ED
+Group: Solid Community Group
+Favicon: https://solidproject.org/TR/solid.svg
+ED: https://solid.github.io/notifications/webpush-subscription-2022
+Repository: https://github.com/solid/notifications
+Inline Github Issues: title
+Markup Shorthands: markdown yes
+Max ToC Depth: 2
+Editor: [Christoph Braun](https://github.com/uvdsl)
+Abstract:
+  The [[!SOLID-NOTIFICATIONS inline]] defines a set of interaction patterns for agents to establish subscriptions to resources in a Solid Storage.
+
+  This specification defines a subscription type that applies these patterns to the [[!PUSH-API inline]].
+Status Text:
+  This section describes the status of this document at the time of its publication.
+
+  This document was published by the [Solid Community Group](https://www.w3.org/community/solid/) as
+  an Editor’s Draft. The sections that have been incorporated have been reviewed following the
+  [Solid process](https://github.com/solid/process). However, the information in this document is
+  still subject to change. You are invited to [contribute](https://github.com/solid/solid-oidc/issues)
+  any feedback, comments, or questions you might have.
+
+  Publication as an Editor’s Draft does not imply endorsement by the W3C Membership. This is a draft
+  document and may be updated, replaced or obsoleted by other documents at any time. It is inappropriate
+  to cite this document as other than work in progress.
+
+  This document was produced by a group operating under the [W3C Community Contributor License Agreement
+  (CLA)](https://www.w3.org/community/about/process/cla/). A human-readable
+  [summary](https://www.w3.org/community/about/process/cla-deed/) is available.
+</pre>
+
+<div class="note">
+This darft is based on a submission for [Web Push Notifications from Solid Pods](https://uvdsl.solid.aifb.kit.edu/conf/2022/icwe/demo).
+<ul>
+<li>The submission's server-side (Pod-side) code can be found [here](https://github.com/uvdsl/solid-web-push).</li>
+<li>The submission's client-side (PWA-side) code can be found [here](https://github.com/uvdsl/solid-web-pwa).</li>
+</ul>
+DISCLAIMER: The submission and its implementation may become obsolete with progression of this draft.
+</div>
+
+# Introduction # {#introduction}
+
+*This section is non-normative.*
+
+The [[!SOLID-NOTIFICATIONS inline]] describes a general pattern by which agents can be notified when a Solid Resource changes.
+
+This document describes a Solid Notifications subscription type that makes use of the [[!PUSH-API inline]] for Web Push notifications in Progressive Web Applications (PWAs).
+
+This specification is for:
+
+* Resource server developers who wish to enable clients, i.e., PWAs, to listen for updates to particular resources.
+* Application developers who wish to implement a client, i.e., a PWA, to listen for updates to particular resources.
+
+## Terminology ## {#terminology}
+
+*This section is non-normative.*
+
+This document uses terminology from the [[!SOLID-NOTIFICATIONS]] protocol, 
+  including "topic".
+
+Issue(62):
+
+Note: Let the predicate `topic` be an `rdfs:subClassOf` of `as:object`?
+
+
+
+The document uses terms from [[!PUSH-API]] specification, 
+  including "push endpoint", "push service" and "authentication secret".
+It also uses terms from [[!OAUTH-2.0]] specification,
+  including  "authorization server" and "access token".
+In addition, the document uses terms from the [[!WEBARCH]] specification, 
+  including "information resource".
+
+This document uses the following terms as defined below:
+: subscription service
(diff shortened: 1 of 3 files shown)


elf-pavlik:
I will review it this weekend, meanwhile, I pushed your branch to my for so that GitHub action renders it. I still need to set up the rendering preview for each PR. Preview of this branch at d5ff597 available here: https://elf-pavlik.github.io/notifications/webpush-subscription-2022

webpush-subscription-2022.bs
```diff
@@ -0,0 +1,360 @@
...
+<ul>
+<li> 
+**Discovery:**
+The *subscriber* discovers from the *storage metadata* a suitable *subscription service*.
+It further discovers from the *subscription service*'s representation a suitable subscription endpoint to submit subscription requests to and the `vapidPublicKey` of the *subscription service*.
+</li>
+<li> 
+**Establish Subscription:**
+The *subscriber* subscribes to the *browser messaging service* to receive Web Push notifications using the `vapidPublicKey` of the *subscription service*.
+In return, the *subscriber* retrieves the `endpoint`, `auth` and `p256dh` values.
```
acoburn, line 112: We're trying to move away from the use of `endpoint`. Depending on the context, we generally use `subscription`, `topic` or `source`. It seems that `subscription` might be best suited here, but I'll leave that up to you.

webpush-subscription-2022.bs
```diff
@@ -0,0 +1,360 @@
...
+</li>
+</ul>
+</div>
+
+# WebPushSubscription2022 Type # {#subscription-type}
+
+This specification defines the WebPushSubscription2022 type for use with Solid Notifications subscriptions.
+The URI of the subscription type is &lt;http://www.w3.org/ns/solid/notification#WebPushSubscription2022&gt;.
+
+An WebPushSubscription2022 API MUST conform to the [Solid Notifications Protocol](https://solid.github.io/notifications/protocol#discovery).
```
acoburn, line 147:
```suggestion
A WebPushSubscription2022 API MUST conform to the [Solid Notifications Protocol](https://solid.github.io/notifications/protocol#discovery).
```

webpush-subscription-2022.bs
```diff
@@ -0,0 +1,360 @@
...
+
+An WebPushSubscription2022 API SHOULD support the [Solid Notifications Features](https://solid.github.io/notifications/protocol#notification-features).
```
acoburn, line 149:
```suggestion
A WebPushSubscription2022 API SHOULD support the [Solid Notifications Features](https://solid.github.io/notifications/protocol#notification-features).
```

webpush-subscription-2022.bs
```diff
@@ -0,0 +1,360 @@
...
+%% : object
+%% :: The `object` property indicates the information resource which the client would like to receive notifications about. 
+%%    The value of source property MUST be a URI, using the `https` scheme, and identifying an information resource.
+
+%% Note: Let the predicate `topic` be an `rdfs:subClassOf` of `as:object`?
+
+%% Issue(62):
+
+: endpoint
+:: The `endpoint` property indicates the "Push Endpoint" as defined in the [[!PUSH-API]] specification.
```
acoburn, line 167: The Push API, does refer to this as an "endpoint", but in our context, `subscription` is likely the better term

webpush-subscription-2022.bs
```diff
@@ -0,0 +1,360 @@
...
+POST /subscribe/
+Authorization: DPoP <token>
+DPoP: <proof>
+Content-Type: text/turtle
+```
+```turtle
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix as: <https://www.w3.org/ns/activitystreams#> .
+@prefix push: <https://purl.org/solid-web-push/vocab#> .
+@prefix notify: <https://www.w3.org/ns/solid/notification> .
```
acoburn, line 207:
```suggestion
@prefix notify: <https://www.w3.org/ns/solid/notification#> .
```

webpush-subscription-2022.bs
```diff
@@ -0,0 +1,360 @@
...
+        "href": "https://solidproject.org/TR/protocol",
+        "title": "Solid Protocol",
+        "publisher": "W3C Solid Community Group"
+    },
+    "SOLID-NOTIFICATIONS": {
+        "authors": [
+            "Aaron Coburn",
+            "Sarven Capadisli"
+        ],
+        "href": "https://solid.github.io/notifications/protocol",
```
acoburn, line 302: not to change now, but just FYI that this will (soon?) be under the `https://solidproject/TR/` URI space

webpush-subscription-2022.bs
```diff
@@ -0,0 +1,360 @@
...
+        "title": "Solid Notifications Protocol",
+        "publisher": "W3C Solid Community Group"
+    },
+    "SOLID-OIDC": {
+        "authors": [
+            "Aaron Coburn",
+            "elf Pavlik",
+            "Dmitri Zagidulin"
+        ],
+        "href": "https://solid.github.io/solid-oidc",
```
acoburn, line 312: not to change now, but just FYI that this will (soon?) be under the `https://solidproject/TR/` URI space

acoburn (commented):
This looks like a great start. I look forward to discussing this more in depth at an upcoming panel meeting

webpush-subscription-2022.bs
uvdsl: Hm. I am hesitant to call the property 'subscription' / `pushSubscription`.
- the property name is already defined by the Push API
- the property is also called `endpoint` in the corresponding JS implementation of the Push API. 
I think renaming this property would make it more confusing to an implementer.

On the other hand,
> A [push subscription](https://www.w3.org/TR/push-api/#dfn-push-subscription) has an associated push endpoint. It MUST be the absolute URL exposed by the [push service](https://www.w3.org/TR/push-api/#dfn-push-service) where the [application server](https://www.w3.org/TR/push-api/#dfn-application-server) can send [push messages](https://www.w3.org/TR/push-api/#dfn-push-message) to. A [push endpoint](https://www.w3.org/TR/push-api/#dfn-push-endpoint) MUST uniquely identify the push subscription.

indicates that the `endpoint` is quite the overloaded term when they try to say that the URL identifies the push subscription.

If I had to switch from `endpoint` to something else, I'd prefer  `pushSubscription` over `subscription` just to be explicit which one is meant.

webpush-subscription-2022.bs
uvdsl: Thank you. I fixed the typo.

webpush-subscription-2022.bs
uvdsl: Thank you. I fixed the typo.

webpush-subscription-2022.bs
uvdsl: Thank you. I fixed the typo.

webpush-subscription-2022.bs
uvdsl: This particular `endpoint` is the "Push Endpoint" of the PUSH API. Discussion in the other comment on code.

However, I did use the term `subscription endpoint` which I now have replaced with `Notification Subscription API` as used by the Notifications Protocol.

webpush-subscription-2022.bs
elf-pavlik: Since this is a property on the subscription request, not the subscription response I would not use `subscription` here. I think other subscription types call this endpoint a `target, since the publisher of notifications, should deliver them to that `target` endpoint.

elf-pavlik (approved):
If we think we need resolve `endpoint` / `target` on subscription **request** as separate issue

csarven (approved):
Approving (and will follow with a merge) as per https://github.com/solid/notifications-panel/blob/main/meetings/2022-05-05.md#define-webpushsubscription2022-type-for-solid-notifications with the understanding that there are no significant blockers or objections for this document to be added here as the initial Editors Draft. The work on the work item can continue - including issues that it needs a solution for, as well as the unresolved conversations here.

Thanks for kicking this off.

Issue #67, opened 2022-05-26
Additional terminology definitions


acoburn:
In the process of linking the vocabulary terms to relevant locations in the specification, I have found that a number of terms that are used in the protocol specification are left undefined. These include:

* `notify:subscription`
* `notify:notificationChannel`
* `notify:feature`

There should be normative definitions for each of these.

Issue #68, opened 2022-05-26
Rename features to feature


acoburn:
The term `features` should be renamed to `feature`, as is customary for RDF predicates.

PR #70, opened 2022-06-02
Add definitions for notification channel discovery and capabilities

csarven:
Resolves https://github.com/solid/notifications/issues/67 and https://github.com/solid/notifications/issues/68

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -491,6 +491,19 @@ content: "";
               <section id="discovery" inlist="" rel="schema:hasPart" resource="#discovery">
                 <h3 property="schema:name">Discovery</h3>
                 <div datatype="rdf:HTML" property="schema:description">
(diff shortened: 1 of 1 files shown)


acoburn:
I'm :+1: on the material portion of this change.

Can we add a "changelog" of sorts to the end of the document outlining these sorts of significant changes? I would expect the changelog is only relevant for pre 1.0 versions. For 1.0, the changelog would be removed.

Inspired by the [Document History sections](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-dpop#appendix-B) you see in IETF specs

acoburn (approved):

csarven:
Followed up in https://github.com/solid/notifications/commit/707888f62d5ff2bd15396d57c3a72163a534442c for this specific change.

Issue #57, opened 2022-03-28
Protocol doesn't mention Subscription Response


elf-pavlik:
https://solid.github.io/notifications/protocol#subscription

This section only describes a payload of Subscription Request. At the same time, various subscription types describe requirements on various Subscriptions Responses.

I think protocol should mention Subscription Response and shortly explain their role and mention that each subscription type will specify details of Subscription Response.

elf-pavlik:
For example https://solid.github.io/notifications/linkeddatanotifications-subscription-2021#subscription-type

> The `inbox` property is used in the body of the subscription request.
> The `webid` property is used in the body of the subscription response.

PR #72, opened 2022-06-02
Add server-subscription-request-response

csarven:
Resolves https://github.com/solid/notifications/issues/57

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -607,21 +607,28 @@ content: "";
                     <dd about="#subscription-response" property="skos:definition">A <em>subscription response</em> is an HTTP response to a <a href="#subscription-request">subscription request</a>. The payload of the subscription response are specialised by <a href="#subscription-types">subscription types</a>.</dd>
                   </dl>
 
-                  <p><span about="" id="server-subscription-resource-safe-methods" rel="spec:requirement" resource="#server-subscription-resource-safe-methods"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> accept <code>GET</code>, <code>HEAD</code> and <code>OPTIONS</code> requests targeting the subscription resource.</span></span></p>
+                  <p><span about="" id="server-subscription-request-methods" rel="spec:requirement" resource="#server-subscription-request-methods"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> support the <code>GET</code>, <code>HEAD</code>, <code>OPTIONS</code> and <code>POST</code> methods [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>] methods on the subscription resource.</span></span></p>
 
                   <p>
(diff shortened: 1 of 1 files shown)


acoburn (approved):

elf-pavlik:
I think that besides the requirement we also should have clear definitions of
* Subscription Request
* Subscription Response

Having that the requirement could clearly state that it adds specific details to either one of those two.

With current PR we still don't have a clear way in subscription type to extend requirements on Subscription Response, since the protocol doesn't define it.

csarven:
Updated in https://github.com/solid/notifications/pull/72/commits/c8892edc434acdad27e8e87e370bf80ec40b4656 .. We'll figure out later whether we move some of this into Terminology or not. I think this is okay for this PR.

protocol.html
```diff
@@ -663,6 +672,8 @@ <h3 property="schema:name">Subscription</h3>
                       </span>
                     </dd>
                   </dl>
+
+                  <p><span about="" id="server-subscription-request-response" rel="spec:requirement" resource="#server-subscription-request-response"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> respond to a successful subscription including the fields: <code>@context</code>, <code>type</code>, in addition to other fields described by specific subscription types.</span></span></p>
```
uvdsl, line 676: Just a minor comment:

> Servers MUST include the field `@context` in a response
may we add: (when serving content-type `application/ld+json`)?

If my client would like to talk turtle with the server and the server supports turtle, why would there be a requirement for `@context`? (how would one do that anyway? genuine question...)
The Notifications Protocol says that serialisations other than JSON-LD should not be prohibited....

csarven: The context is clear but I updated in https://github.com/solid/notifications/pull/72/commits/c68ed9e7ff55f993eade5828894efbc303b38dc1 to mention JSON-LD payload.

Server may mention the "accept" feature in the notification channel description and client can use that in the request: https://solid.github.io/notifications/protocol#feature-accept

uvdsl: The "accept" feature is for notifications, right?
But the subscription response MUST be in JSON-LD? 
My understanding is that JSON-LD is the minimum serialisation required, but any other may be supported as well.
This holds also for subscription request and response. Correct me if I am wrong, though.

So, I thought if it shouldn't be phrased less restrictive in the sense, that if the server supports receiving a request in turtle and sending the response in turtle, then it should also work?

uvdsl: As discussed after the meeting last week (7 July 2022), we could clarify that the response is subject to content negotiation?
And thus, if JSON-LD is used, `@context` must be included, additionally to the generally required `type`.

csarven: Okie dokie. The changes in https://github.com/solid/notifications/pull/72/commits/29ad87f865f734f765b52d2f799cef3dfe50e00e are intended to clarify server's:
* support of HTTP methods targeting the subscription resource,
* accepting request payloads in JSON-LD, and
* responding to requests preferring a JSON-LD payload.

csarven:
Merging as per https://github.com/solid/notifications-panel/pull/68/files#diff-697e56f4b6e7271430f73e456bd8dc14cec4e0332eb60255b77dfb9088374028R59 (the minutes are under review as of this writing).

PR #99, opened 2022-08-22
Remove server-subscription-access-read

csarven:
This requirement will prohibit use cases categorically requiring notifications about resources that do not require clients to have read access on the resource.

Systems may want to enable clients to be notified when a specific resource is created or access controls on the resource are changed. That is orthogonal to systems disallowing subscriptions to a specific resource (which may or may not have a representation), for any reason (at any time), including those in which the client does not have read access to the resource.

Systems may want to advertise and apply subscription policies.


acoburn (approved):
This is a very useful generalization

csarven:
Merged as ACTION of https://github.com/solid/notifications-panel/blob/main/meetings/2022-08-25.md#remove-server-subscription-access-read

Issue #58, opened 2022-04-01
Discovery of Notification Subscription Metadata


csarven:
The Notifications Protocol currently does not describe the discovery of the Notification Subscription Metadata Resource.

This issue is a follow-up on the discussion at https://github.com/solid/notifications/pull/54 . Some options are mentioned there in addition to our earlier consideration to use Server/Storage Metadata: https://github.com/solid/specification/issues/355

csarven:
Some considerations:

To clarify, https://github.com/specification/issue/355 is of interest in the context of discovery of notification channels that's available for a given resource starts from a resource of interest, finds the storage the resource is included in, and that the resource server provides channel information for all of its resources in the storage description (wherever that's exactly materialised) - where the resource of interest for notifications will be covered by.

While that may be sufficient for most use - and completely okay to just stop there - it is worthwhile to clarify here whether Notifications Protocol should describe how any resource can advertise its own notification channels. So not all resources in the same storage will necessarily use the same channel e.g., a particular container or non-container resource could call its own. I don't have a strong use cases in mind for this but then again it is more intuitive than the channels being offered for all resources in a storage. The fact that we are talking about storage-level channels is only because of Solid Protocol and the concept of 'storage' existing. Whether the concept of 'storage' is fundamental to Notifications Protocol is yet another question. In practical terms, can/should the Notifications Protocol be usable alongside protocols besides Solid Protocol? (I'd love it to be the case if we can make it so.)

There is some consensus on providing the mechanism to discover storage description - from the perspective of Editors Team [2022-06-08 meeting](https://github.com/solid/specification/blob/main/meetings/2022-06-08.md). We'll work out some of the finer details in issue 355 and followup.

On a related note, the discovery of a resource's ACL resources, using `acl` link relation, is defined in the WAC spec, as opposed to the Solid Protocol. So, again, the question of whether Notifications Protocol should describe its own discovery for notification channels.

I can also see it being fine for the Notifications Protocol to only specify what the channel description should be and leave out its discovery to other protocols to specify.

uvdsl:
re: Should a resource be able to advertise a non-storage-bound notification channel?

As an example use-case, let's say that there is a `storage A` that does support LDNs and Websockets but does not support Web Push notifications. And on `storage S` there lives a `resource r`. A user would like to receive Web Push notifications on that resource. (Web Push just for the sake of the argument, you can also have some other notification type that is not supported by `storage S`) But since the storage does not support Web Push, some other service is needed.

[...]

PR #104, opened 2022-08-25
Update notification channel discovery and conformance

csarven:
Resolves issue #58 by working with the notion of notification channel:
* Adds definition for notification-channel-resource.
* Describes notification-channel-discovery via notification-channel-resource (servers can advertise link relations `describedby` ( server-link-resource-notification-channel ) and `http://www.w3.org/ns/solid#storageDescription` ( server-link-storage-notification-channel ) and notification-channel description.
* Adds notification-channel-conformance

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -410,22 +410,22 @@ content: "";
 
                   <p property="skos:definition">This specification defines the following terms. These terms are referenced throughout this specification.</p>
 
-                  <span rel="skos:hasTopConcept"><span resource="#notification-subscription-metadata-resource"></span><span resource="#subscription-resource"></span><span resource="#subscription-request"></span><span resource="#subscription-response"></span></span>
+                  <span rel="skos:hasTopConcept"><span resource="#notification-channel"></span><span resource="#notification-channel-resource"></span><span resource="#subscription-resource"></span><span resource="#subscription-request"></span><span resource="#subscription-response"></span></span>
 
                   <dl>
                     <dt about="#notification-channel" property="skos:prefLabel" typeof="skos:Concept">
-                      <dfn id="notification-channel">Notification Channel</dfn>
+                      <dfn id="notification-channel">notification channel</dfn>
                     </dt>
                     <dd about="#notification-channel" property="skos:definition">
-                      A <em>Notification Channel</em> is an abstract thing which is identified by a URI and whose properties describe the available communication interface and features.
+                      A <em>notification channel</em> is an abstract thing which is identified by a URI and whose properties describe the available communication interface and features.
                     </dd>
 
-                    <dt about="#notification-subscription-metadata-resource" property="skos:prefLabel" typeof="skos:Concept">
-                      <dfn id="notification-subscription-metadata-resource">Notification Subscription Metadata Resource</dfn>
+                    <dt about="#notification-channel-resource" property="skos:prefLabel" typeof="skos:Concept">
+                      <dfn id="notification-channel-resource">notification channel resource</dfn>
                     </dt>
-                    <dd about="#notification-subscription-metadata-resource" property="skos:definition">
+                    <dd about="#notification-channel-resource" property="skos:definition">
                       An <em>RDF document</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>]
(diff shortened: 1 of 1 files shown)


protocol.html
```diff
@@ -509,59 +509,62 @@ <h3 property="schema:name"><span>Issue</span>: Solid Notification Vocabulary and
...
-                    <div datatype="rdf:HTML" property="schema:description">
-                      <p>The specific mechanism defining the discovery of the <a href="#notification-subscription-metadata-resource">Notification Subscription Metadata Resource</a> is under consideration: <a href="https://github.com/solid/notifications/issues/58" rel="cito:citesAsRelated">issues/58</a>.</p>
+                      <p>
+                        <span about="" id="server-notification-channel-resource-accept" rel="spec:requirement" resource="#server-notification-channel-resource-accept">
+                          <span property="spec:statement">
+                            <span rel="spec:requirementSubject" resource="spec:Server">Servers</span>
+                            <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> accept requests on the Notification Channel Resource when the value of the <code>Accept</code> header indicates a preferred representation in <code>application/ld+json</code> [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>].
```
acoburn, line 523: Is it necessary to require JSON-LD 1.1, or can we just say JSON-LD?

protocol.html
```diff
@@ -509,59 +509,62 @@ <h3 property="schema:name"><span>Issue</span>: Solid Notification Vocabulary and
...
+                    <h4 property="schema:name">Notification Channel Discovery</h4>
+                    <div datatype="rdf:HTML" property="schema:description">
+                      <p about="" id="server-link-resource-notification-channel" rel="spec:requirement" resource="#server-link-resource-notification-channel"><span property="spec:statement">When a server wants to enable applications to discover <a href="#notification-channel">Notification Channels</a> available to a given <a href="https://solidproject.org/TR/protocol#resource">resource</a>, the <span rel="spec:requirementSubject" resource="spec:Server">server</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> advertise the associated resources describing the channel information by responding to an HTTP request including a <code>Link</code> header with the <code>rel</code> value of <code>describedby</code> [<cite><a class="bibref" href="#bib-powder-dr">POWDER-DR</a></cite>] and the Description Resource as link target [<cite><a class="bibref" href="#bib-rfc8288">RFC8288</a></cite>].</span></p>
```
acoburn, line 515: Powder ([appendix D](https://www.w3.org/TR/2009/REC-powder-dr-20090901/#appD)) puts no constraints on the format of the target of a `describedby` link relation (it could be an image). Should this requirement add a constraint such that a description be at least content-negotiatable as RDF?

acoburn (approved):
This looks good, just a few clarifying questions

protocol.html
csarven: I think JSON-LD will suffice and generally entails the latest published version.

But this reminds me to mention something related about the JSON-LD form. Should mention requiring https://www.w3.org/TR/json-ld11/#compacted-document-form in favour of the details on the context value being array currently.

protocol.html
csarven: True. I think this is explained further in notification-channel-resource (RDF Document). Would that suffice?

protocol.html
csarven: We know the spec works with JSON-LD11 so we should require and reference that specifically.

I will update the references to JSON-LD11 throughout the document separately - not part of this PR.

I can also follow-up on any changes to context and document-form in a separate PR. Discussion here or elsewhere can continue.

protocol.html
acoburn: That would suffice for me

protocol.html
acoburn: > Should we require https://www.w3.org/TR/json-ld11/#compacted-document-form instead of @context with an array?

A document with `"@context": [ ... ]` _is_ compacted document form

acoburn:
I'm +1 to merge in the current form

csarven:
Follows https://github.com/solid/notifications-panel/blob/main/meetings/2022-08-25.md#update-notification-channel-discovery-and-conformance

Issue #77, opened 2022-06-22
Clarify Notifications Flow diagrams


csarven:
Diagram should be clarified (content and visuals) to better distinguish between what's defined by the Notifications Protocol from the components/interactions defined elsewhere.

Issue #106, opened 2022-09-02
Notifications Protocol Conformance Classes


csarven:
Add a Conformance Classes section listing the roles eg. Notification Server, Client, that the Solid Notifications Protocol specifies. The criteria for each of the roles can be described throughout the specification.

elf-pavlik:
I proposed some new terminology in #108 

We already have #49 and I recall a discussion that we should minimize assumptions on how Subscriber and Publisher internally choose to couple or separate different responsibilities.


I'd look separately at the Subscription Flow and Notifications Flow. The first seems a bit simpler since the client/server roles match the subscriber/publisher. I think we have distinct:

* Subscriber - client sending subscription request
* Topic publisher - server publishing the topic
  * Storage description
  * Subscription channel 

I think we shouldn't be careful how we look at the publisher side in the subscription flow, probably we can consider it as a single entity even though internally the implementations may choose to separate topics publisher from subscription channels.

Now for the Notifications flow, depending on the source (pick-up) or target (delivery) flow the client/server roles will work in opposite way:

* source (pick-up)
  * subscriber - client
  * publisher - server
* target (delivery)
  * subscriber - server
  * publisher - client

We probably also externally want to just look at Subscriber and Publisher leaving out all the coupling or separation related to notifications source and notifications target as their implementation details.

PR #108, opened 2022-09-08
Simplify flow diagrams

acoburn:
This simplifies the flow diagrams, removing those actors that are out of scope for this specification and placing the common portions of the flow into its own diagram. And given that `source` and `target` are not defined until later in the document, this uses more generally understood terminology to describe the categories of interaction.

PR towards https://github.com/solid/notifications/issues/77

### DIFF
--- a/notifications-flow-target.mmd
+++ b/notifications-flow-target.mmd
@@ -4,6 +4,6 @@ sequenceDiagram
   participant Notifications Target
 
   loop for each notification
-    Notifications Publisher ->> Notifications Target: Deliver notification
-    Notifications Target ->> Notifications Publisher: 200 OK
+    Notifications Sender ->> Notifications Target: Deliver notification
+    Notifications Target ->> Notifications Sender: 200 OK
   end
--- a/notifications-flow-target.mmd
+++ b/notifications-flow-target.mmd
@@ -1,6 +1,6 @@
 sequenceDiagram
   autonumber
-  participant Notifications Publisher
+  participant Notifications Sender
   participant Notifications Target
 
   loop for each notification
--- a/notifications-flow-target.mmd
+++ b/notifications-flow-target.mmd
@@ -1,12 +1,9 @@
 sequenceDiagram
   autonumber
-  participant Subscriber
+  participant Notifications Publisher
   participant Notifications Target
-  participant Subscription Resource
-  participant Notifications Source
 
   loop for each notification
-    Notifications Source ->> Notifications Target: Deliver notification
+    Notifications Publisher ->> Notifications Target: Deliver notification
+    Notifications Target ->> Notifications Publisher: 200 OK
   end
-  Subscriber ->> Subscription Resource: Unsubscribe (delete subscription)
-  Subscription Resource ->> Subscriber: 204 No Content
--- a/notifications-flow-target.mmd
+++ b/notifications-flow-target.mmd
@@ -2,19 +2,11 @@ sequenceDiagram
   autonumber
   participant Subscriber
   participant Notifications Target
-  participant Storage Metadata Resource
-  participant Subscriptions Container
-  participant Authorization Server
+  participant Subscription Resource
   participant Notifications Source
 
-  Subscriber ->> Storage Metadata Resource: Discovery
-  Storage Metadata Resource ->> Subscriber: Storage Metadata Resource
-  Subscriber ->> Subscriptions Container: Request subscription (with access token)
-  Subscriptions Container ->> Authorization Server: Verify authorization
-  Authorization Server ->> Subscriptions Container: capabilities
-  Subscriptions Container ->> Subscriber: Subscription response (with webid & unsubscribe)
   loop for each notification
     Notifications Source ->> Notifications Target: Deliver notification
   end
-  Subscriber ->> Subscriptions Container: Unsubscribe (delete subscription)
-  Subscriptions Container ->> Subscriber: 204 No Content
+  Subscriber ->> Subscription Resource: Unsubscribe (delete subscription)
+  Subscription Resource ->> Subscriber: 204 No Content
--- a/protocol.html
+++ b/protocol.html
@@ -407,7 +407,7 @@ content: "";
                   <figure id="solid-notifications-flow-target" rel="schema:hasPart" resource="#solid-notifications-flow-target">
                     <img src="notifications-flow-target.mmd.svg" rel="schema:image" width="800" />
 
-                    <figcaption property="schema:name">Solid Notifications Flow, where the publisher connects to the notification target</figcaption>
+                    <figcaption property="schema:name">Solid Notifications Flow, where the sender delivers to the notification target</figcaption>
                   </figure>
                 </div>
               </section>
--- a/protocol.html
+++ b/protocol.html
@@ -399,7 +399,7 @@ content: "";
                   <figure id="solid-notifications-flow-source" rel="schema:hasPart" resource="#solid-notifications-flow-source">
                     <img src="notifications-flow-source.mmd.svg" rel="schema:image" width="800" />
 
-                    <figcaption property="schema:name">Solid Subscription Flow for Connected Clients</figcaption>
+                    <figcaption property="schema:name">Solid Subscription Flow, where the subscriber connects to the notification source</figcaption>
                   </figure>
 
                   <p>Subscription types in which a client establishes a subscription and does not maintain a persistent connection to a Notifications Source are illustrated by the following diagram.</p>
@@ -407,7 +407,7 @@ content: "";
                   <figure id="solid-notifications-flow-target" rel="schema:hasPart" resource="#solid-notifications-flow-target">
                     <img src="notifications-flow-target.mmd.svg" rel="schema:image" width="800" />
 
-                    <figcaption property="schema:name">Solid Notifications Flow for Offline Clients</figcaption>
+                    <figcaption property="schema:name">Solid Notifications Flow, where the publisher connects to the notification target</figcaption>
                   </figure>
                 </div>
               </section>
--- a/protocol.html
+++ b/protocol.html
@@ -385,19 +385,29 @@ content: "";
                 <div datatype="rdf:HTML" property="schema:description">
                   <p><em>This section is non-normative.</em></p>
 
-                  <p>The following diagrams show the high-level interactions involved in <em>source</em> and <em>target</em> flows.
+                  <p>The following diagram shows the high-level interactions involved discovering and establishing a notification subscription.
                      How a client retrieves an access token for step 3 is outside the scope of this document.</p>
+
+                  <figure id="solid-notifications-flow-protocol" rel="schema:hasPart" resource="#solid-notifications-flow-protocol">
+                    <img src="notifications-flow-protocol.mmd.svg" rel="schema:image" width="800" />
+
+                    <figcaption property="schema:name">Solid Notifications Flow</figcaption>
+                  </figure>
+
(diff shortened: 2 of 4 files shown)


protocol.html
```diff
@@ -384,19 +384,29 @@ <h3 property="schema:name">Overview</h3>
...
+                    <figcaption property="schema:name">Solid Subscription Flow for Connected Clients</figcaption>
                   </figure>
 
+                  <p>Subscription types in which a client establishes a subscription and does not maintain a persistent connection to a Notifications Source are illustrated by the following diagram.</p>
+
                   <figure id="solid-notifications-flow-target" rel="schema:hasPart" resource="#solid-notifications-flow-target">
                     <img src="notifications-flow-target.mmd.svg" rel="schema:image" width="800" />
 
-                    <figcaption property="schema:name">Solid Notifications Flow (target)</figcaption>
+                    <figcaption property="schema:name">Solid Notifications Flow for Offline Clients</figcaption>
```
elf-pavlik, line 409: I don't think *Offline Clients* makes sense here.
Actually, client-server terminology might be confusing since the Subscriber acts as the server when a notification gets delivered, and the notifications publisher will act as a client.

Actually, the Subscriber needs to maintain their `target` online to be able to receive incoming requests with notifications.

elf-pavlik (changes requested):

protocol.html
elf-pavlik:
```suggestion
                    <figcaption property="schema:name">Solid Notifications Flow, where the publisher connects to the notification target (from the <a href="#subscription-request">subscription request</a>)</figcaption>                    
```

again `target` chan be changed in dedicated PR

elf-pavlik:
I think separating
* Subscription Flow
* Notifications Flow

into separate diagrams is very helpful!

elf-pavlik:
Another option: 

```mermaid
sequenceDiagram
  autonumber
  participant Notifications Target
  participant Subscriber
  participant Storage Description Resource
  participant Subscription Resource

  Subscriber ->> Storage Description Resource: Discovery
  Storage Description Resource ->> Subscriber: Storage Description Resource
  Subscriber ->> Subscription Resource: Request subscription (with access token)
  Subscription Resource ->> Subscriber: Subscription response (with subscription information)

  alt pick-up
    Subscriber ->> Notifications Source: Establish connection
    Notifications Source -->> Subscriber: Stream notifications
  else
    loop for each notification
      Notifications Source ->> Notifications Target: Deliver notification
      Notifications Target ->> Notifications Source: 200 OK
    end
  end

  Subscriber ->> Subscription Resource: Unsubscribe (delete subscription)
  Subscription Resource ->> Subscriber: 204 No Content
```

elf-pavlik:
I think separate

* Subscription Flow (with Unsubscribe)
* Notification Flow

Would be the cleanest approach.

```mermaid
sequenceDiagram
  autonumber
  participant Subscriber
  participant Storage Description Resource
  participant Subscription Resource

  Subscriber ->> Storage Description Resource: Discovery
  Storage Description Resource ->> Subscriber: Storage Description Resource
  Subscriber ->> Subscription Resource: Subscription Request (with access token)
  Subscription Resource ->> Subscriber: Subscription Response (with subscription information)

  Note over Subscriber, Subscription Resource: Notifications Flow

  Subscriber ->> Subscription Resource: Unsubscribe (delete subscription)
  Subscription Resource ->> Subscriber: 204 No Content
```

```mermaid
sequenceDiagram
  autonumber
  participant Subscriber
  participant Notifications Source

  Subscriber ->> Notifications Source: Establish connection
  Notifications Source -->> Subscriber: Stream notifications
```

```mermaid
sequenceDiagram
  autonumber
  participant Notifications Target
  participant Notifications Publisher

  loop for each notification
    Notifications Publisher ->> Notifications Target: Deliver notification
    Notifications Target ->> Notifications Publisher: 200 OK
  end
```

In the implementation I'm working on, which implements the subscriber part, I have separate background job queues for subscribing and handing incoming notifications. There are separate workers handling subscriptions and notifications which could eventually run as independent microservices.

elf-pavlik:
I think we should add an inline issue referencing #106 and once it gets resolved update all affected naming.

acoburn:
@elf-pavlik I have updated the diagrams and terminology to be in line with what was discussed in the 10 Nov meeting

elf-pavlik (approved):
LGTM

notifications-flow-target.mmd
```diff
@@ -1,20 +1,9 @@
 sequenceDiagram
   autonumber
-  participant Subscriber
+  participant Notifications Publisher
```
csarven, line 3:
```suggestion
  participant Notifications Sender
```

notifications-flow-target.mmd
```diff
@@ -1,20 +1,9 @@
...
-  Subscriber ->> Storage Metadata Resource: Discovery
-  Storage Metadata Resource ->> Subscriber: Storage Metadata Resource
-  Subscriber ->> Subscriptions Container: Request subscription (with access token)
-  Subscriptions Container ->> Authorization Server: Verify authorization
-  Authorization Server ->> Subscriptions Container: capabilities
-  Subscriptions Container ->> Subscriber: Subscription response (with webid & unsubscribe)
   loop for each notification
-    Notifications Source ->> Notifications Target: Deliver notification
+    Notifications Publisher ->> Notifications Target: Deliver notification
+    Notifications Target ->> Notifications Publisher: 200 OK
```
csarven, line 8:
```suggestion
    Notifications Sender ->> Notifications Target: Deliver notification
    Notifications Target ->> Notifications Sender: 200 OK
```

protocol.html
```diff
@@ -385,19 +385,29 @@ <h3 property="schema:name">Overview</h3>
...
+                    <figcaption property="schema:name">Solid Subscription Flow, where the subscriber connects to the notification source</figcaption>
                   </figure>
 
+                  <p>Subscription types in which a client establishes a subscription and does not maintain a persistent connection to a Notifications Source are illustrated by the following diagram.</p>
+
                   <figure id="solid-notifications-flow-target" rel="schema:hasPart" resource="#solid-notifications-flow-target">
                     <img src="notifications-flow-target.mmd.svg" rel="schema:image" width="800" />
 
-                    <figcaption property="schema:name">Solid Notifications Flow (target)</figcaption>
+                    <figcaption property="schema:name">Solid Notifications Flow, where the publisher connects to the notification target</figcaption>
```
csarven, line 410:
```suggestion
                    <figcaption property="schema:name">Solid Notifications Flow, where the sender delivers to the notification target</figcaption>
```

csarven (approved):
I understand the intention is about creating and distributing the notification and "publisher" suffices. Just want to note that publishing may be generally associated with publishing (and serving) a document on a Web site.

"Sender" (or "producer") of content may also work. "Sender" (alongside "recipient) is also terminology that's common/used in HTTP and LDN (FWIW).

Approving so that we move the work forward and not get stuck on this. We do however need to revisit all terminology, especially in the context of JSON-LD and the Notify vocabulary: https://github.com/solid/vocab/pull/62 (and all the issues touching it / touched by it).

elf-pavlik:
> Just want to note that publishing may be generally associated with publishing (and serving) a document on a Web site.

Besides publishing the notifications, the *publisher* also publishes the *topic* to which subscriber wants to subscribe to.

I'm open to further iterations on the terminology and fully agree that this PR as it stands makes a good step forward.

csarven:
>Besides publishing the notifications, the publisher also publishes the topic to which subscriber wants to subscribe to.

I don't quite follow the context of that sentence. What's the significance of "topic" besides it being a unit of information in a notification? Is the "publisher" intended to manage the "topic" resource that's of interest to the subscriber? As I understand it, the "publisher" is a role of creating/producing/sending/delivering/distributing a notification, and that the publication/having a representation of the topic resource is done by some resource server.

notifications-flow-target.mmd
```diff
@@ -1,20 +1,9 @@
...
-  Storage Metadata Resource ->> Subscriber: Storage Metadata Resource
-  Subscriber ->> Subscriptions Container: Request subscription (with access token)
-  Subscriptions Container ->> Authorization Server: Verify authorization
-  Authorization Server ->> Subscriptions Container: capabilities
-  Subscriptions Container ->> Subscriber: Subscription response (with webid & unsubscribe)
   loop for each notification
-    Notifications Source ->> Notifications Target: Deliver notification
+    Notifications Sender ->> Notifications Target: Deliver notification
+    Notifications Target ->> Notifications Sender: 200 OK
   end
```
csarven, line 9: If `loop` is removed:

```suggestion
```

csarven (changes requested):
Requesting changes mainly for the mention of removing "access token".

notifications-flow-target.mmd
```diff
@@ -1,20 +1,9 @@
...
-  participant Authorization Server
-  participant Notifications Source
 
-  Subscriber ->> Storage Metadata Resource: Discovery
-  Storage Metadata Resource ->> Subscriber: Storage Metadata Resource
-  Subscriber ->> Subscriptions Container: Request subscription (with access token)
-  Subscriptions Container ->> Authorization Server: Verify authorization
-  Authorization Server ->> Subscriptions Container: capabilities
-  Subscriptions Container ->> Subscriber: Subscription response (with webid & unsubscribe)
   loop for each notification
```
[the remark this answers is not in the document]
acoburn: No strong opinion on this. I suspect that the loop construct could be omitted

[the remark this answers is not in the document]
elf-pavlik: Each notification results in separate request to and response from the target, in contrast to connecting to the source where all notifications are streamed in the same ongoing response.

If you look at https://github.com/solid/notifications/pull/108#issuecomment-1272336602 subscription flow  has a block where notifications flow would conceptually fit in. We still have open question of adding unsubscribe and once (if) we add it, having single request+response to the target between subscribe and unsubscribe doesn't illustrate how it works very well.

In short, the diagrams are non-normative and for the sake of illustrating how things work, the loop IMO comes helpful.

elf-pavlik:
>  As I understand it, the "publisher" is a role of creating/producing/sending/delivering/distributing a notification, and that the publication/having a representation of the topic resource is done by some resource server.

How publisher is deploying there services is out of scope. Still both the resource which is referenced as the *topic* of the subscription and the party which is delivering notifications to the target, or providing the notifications source, are working together. Since we don't define any details how the resource server and the notifying party are working together, they could be part of the same application or separate services using implementation specific way of communicating.
Actually the subscription manager could also be a separate service which coordinates with other two mentioned roles. This way we have 3 distinct responsibilities under a single umbrella of the *publisher*:

* resource server hosting the *topic*
* subscription manager - we define how subscriber can discover it
* notifying party - we only define how subscriber either fetches notifications for source or exposes target to which notifications get delivered

This conversation should leave this PR BTW so that we can close it.

csarven:
The context of the interaction here is specifically about a sender delivering a notification to a target. I have no objection to using general terms when necessary, and not in place of more specific terms. I want to know if you have an objection to using sender so we can close the PR.

csarven (approved):

notifications-flow-target.mmd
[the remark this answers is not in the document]
csarven: Perhaps +0.5 for simplified (dropping loop) but no objection to leave loop in.

csarven:
Shouldn't the resource that advertises the discovery (notification channel) resource be included in the diagram in https://solid.github.io/notifications/protocol#solid-notifications-flow-protocol ?

PR #113, opened 2022-10-03
removed custom webhooks authn

elf-pavlik:
I'm starting it as a draft since it needs some more work.

@jaxoncreed do you think we can break up this document into a spec and a primer?
* spec - would intentionally leave AuthN out of scope, only mentioning Solid-OIDC in a non-normative way.
* primer - would show all the details and use specifically Solid-OIDC to provide complete guidance.

### TODO

* [ ] adjust conformance classes based on #120


elf-pavlik:
#146 defines WebhookChannel2023

Once we merge it we can write a primer to accompany it.

PR #117, opened 2022-11-16
Add consider-exposing-information

csarven:


CxRes:
I would suggest you split these up into separate paragraphs for server and client. Also, you probably don't want to use "server" and "client" to be consistent.

elf-pavlik (approved):

PR #118, opened 2022-11-16
Add consider-subscription-resource-trust

csarven:


csarven:
"authoritiesm" was a typo of "authorities". In any case, I removed the mention of discovered through because it is not the most important part - after all, it is really about the subscription resource.

CxRes (approved):

elf-pavlik (approved):
I will follow up with an issue to add specific consideration about retrying requests.

Issue #105, opened 2022-08-26
JSON-LD structural requirements have issues


RubenVerborgh:
The current specification text reads:

> When serializing the Notification Subscription Metadata Resource as JSON-LD, the @contextfield [sic] MUST be an array containing the value https://www.w3.org/ns/solid/notification/v1.

It is unclear whether this was done for developer accessibility, security, and/or other reasons.

However, there are some issues—regardless of whether or not mandating a specific context is a desired solution. I am expanding on this argument also because it can help us inform how other specification documents are written for Solid.

### Assuming that mandating a specific context is desired
1. **The contract is not explicit in HTTP**, as the  context assumption exists as an out-of-band decision.
So clients cannot specifically say they want "the Notifications kind of JSON-LD", and servers cannot indicate that they have responded with it. In other words, [the media type underspecifies the contents](https://ruben.verborgh.org/articles/fine-grained-content-negotiation/#underspecification-at-the-syntactical-level). And while clients could identify the requirement by inspecting the body, such content sniffing is not recommended and causes overhead. A solution here is to explicitly indicate the `profile` of the content (and possibly to [negotiate](https://datatracker.ietf.org/doc/html/draft-svensson-profiled-representations-01#section-4)) it.

2. **The fact that a certain context is present, does not provide any other structural guarantees.**
I am assuming here that the requirement for a specific context was to make the JSON-LD more "developer-friendly", by adding predictability to the JSON structure. However, mandating a specific context does not achieve that, and as such is thus not meaningful by itself. It's easy to think of dozens of different structures to have a JSON-LD document that contains the same triples as the example in the specification, with the same context. So developers cannot usefully derive anything from the context being mandated. A solution here is to require a specific JSON-LD frame instead.

3. **The type field requirement is ambiguous.**
Given the previous point, the `MUST materialize a type field` also has a wide variety of ways to be expressed. The solution is a frame.

4. **Mandating a specific context, and only that context, hinders extensibility.**
Plus, apart from the structural points above, specifying a context does not mandate its usage for abbreviation. So instead of writing `"subscription"`, I can still write "`https://www.w3.org/ns/solid/notification/v1#subscription"` (or whatever its expansion) and have a JSON-LD document that conforms to the Notifications spec.

5. **JSON-LD framing is expensive for the server.**
[...]

PR #120, opened 2022-11-21
Update conformance model

csarven:
This PR develops the under-specified `#conformance` model. Resolves https://github.com/solid/notifications/issues/106 .

The work herein follows the guidelines in https://www.w3.org/TR/spec-variability/ and https://www.w3.org/TR/qaframe-spec/ , and applies http://www.w3.org/ns/spec# .

To simplify the review process, this PR focuses on the `#conformance` section and leaves out updates intended to be throughout the specification corresponding to these changes that will be included in a follow up PR.

PR #122, opened 2022-11-22
Remove specific constraint of JSON-LD compacted-document-form

csarven:
The primary purpose of this PR is to set a uniform data format in the Solid Notifications Protocol based on the data in current requirements (table below). The PR resolves https://github.com/solid/notifications/issues/105 .

|Requirement|`media-type`|Constraints|
|-|-|-|
|`#server-notification-channel-resource-accept`|`application/ld+json`|-|
|`#server-notification-channel-resource-jsonld-context`|`application/ld+json`|Specific serialization of `dfn-compacted-document-form`|
|`#server-subscription-request-accept`|`application/ld+json`|-|
|`#server-subscription-request-content-type`|`application/ld+json`|-|
|`#server-subscription-request-response`|`application/ld+json`|Specific serialization of `dfn-compacted-document-form`|
|`#notification-data-model`|`application/ld+json`|`dfn-compacted-document-form`|

The data shows that while servers (resource server and notification server) are capable of accepting any JSON-LD serialization, the servers are required to materialise a JSON-LD compacted document form or a specific serialization. Clients are expected to produce and consume any JSON-LD serialization.

The rationalise of this PR is to simplify the requirements without breaking code changes or introducing specific code that may not be part of off the shelf JSON-LD libraries by removing the additional constraint that is a specific serialization of the JSON-LD compacted document form.

The PR removes:
* `#server-notification-channel-resource-jsonld-context`
* `#server-subscription-request-response`
* `#subscription-request-context`

Adds:
* `#json-ld-context`

---

As noted elsewhere, there are legitimate reasons to use a "light" data format, e.g., by requiring only JSON-LD compacted document form, a specific serialization of JSON-LD compacted document form, or maybe even something else entirely. That can be pursued by revisiting the requirements as a whole for both servers and clients in addition to advisements.

I suggest that the next change should at least consider setting JSON-LD compacted document form as the required format for both clients and servers.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -500,6 +500,8 @@ content: "";
                 subscription requirements of the notification protocol.
               </p>
 
+              <p id="json-ld-context">In JSON-LD representations, <code>https://www.w3.org/ns/solid/notification/v1</code> can be used as the <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD Context</a>.</p>
+
               <div class="issue" id="solid-notification-vocabulary-context" inlist="" rel="schema:hasPart" resource="#solid-notification-vocabulary-context">
                 <h3 property="schema:name"><span>Issue</span>: Solid Notification Vocabulary and Context</h3>
                 <div datatype="rdf:HTML" property="schema:description">
@@ -540,8 +542,6 @@ content: "";
                         <dt about="#notify-type" property="skos:prefLabel" typeof="skos:Concept"><dfn id="notify-type">type</dfn></dt>
                         <dd about="#notify-type" property="skos:definition"><code>type</code> predicate (<code>rdf:type</code>) denotes the notification channel type (<a href="#notification-channel" rel="rdfs:seeAlso">Notification Channel</a>).</dd>
                       </dl>
-
-                      <p><span about="" id="server-notification-channel-resource-jsonld-context" rel="spec:requirement" resource="#server-notification-channel-resource-jsonld-context"><span property="spec:statement"><span rel="spec:requirementSubject" resource="spec:Server">Servers</span> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> include the <code>@context</code> field as an array containing the value <code>https://www.w3.org/ns/solid/notification/v1</code> in the JSON-LD representation of a Notification Channel Resource.</span></span></p>
                     </div>
                   </section>
 
@@ -644,9 +644,6 @@ content: "";
                   <p id="subscription-request-statements" about="#subscription-request-statements" typeof="skos:Collection"><span property="skos:prefLabel">Subscription request statements</span> include the properties:</p>
 
                   <dl about="#subscription-request-statements" rel="skos:member">
-                    <dt about="#subscription-request-context" property="skos:prefLabel" typeof="skos:Concept"><dfn id="subscription-request-context">@context</dfn></dt>
-                    <dd about="#subscription-request-context" property="skos:definition">An array containing at least the value <code>https://www.w3.org/ns/solid/notification/v1</code> in a JSON-LD payload.</dd>
-
                     <dt about="#subscription-request-type" property="skos:prefLabel" typeof="skos:Concept"><dfn id="subscription-request-type">type</dfn></dt>
                     <dd about="#subscription-request-type" property="skos:definition">A class whose URI indicating the subscription type to be created.</dd>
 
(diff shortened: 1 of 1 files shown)


acoburn (changes requested):
I agree that achieving consistency here is important, but this change goes in the opposite direction of achieving interoperability.

If we intend to support JSON-LD, it is important to align on a particular `@context` IRI. The reason here is that it cannot be assumed that production systems will dereference arbitrary context values (this is a major security risk). If we require a particular context value, it allows us to both support JSON-LD and also the corresponding interoperability.

I would go in a slightly different direction with this PR; instead, I would propose using specific `profile` values in the references to `Accept` headers. IOW, I would suggest that we use:

```
Accept: application/ld+json; profile="https://www.w3.org/ns/solid/notification/v1"
```

This will allow clients to indicate to servers that they support this particular flavor of constrained JSON-LD (as opposed to _any_ JSON-LD serialization)

protocol.html
```diff
@@ -500,6 +500,8 @@ <h2 property="schema:name">Protocol</h2>
                 subscription requirements of the notification protocol.
               </p>
 
+              <p id="json-ld-context">In JSON-LD representations, <code>https://www.w3.org/ns/solid/notification/v1</code> can be used as the <a href="https://www.w3.org/TR/json-ld11/#the-context">JSON-LD Context</a>.</p>
```
RubenVerborgh, line 503: `MAY`?

RubenVerborgh (approved):

csarven:
Updated in https://github.com/solid/notifications/pull/122/commits/b73a814d5baf9dd22cb54adbf5b546afad7c1bf2 . Hopefully that brings us closer to what's intended. Will get back to re-structuring the content and markup after this PR - partly due to resolving other PRs.

protocol.html
```diff
@@ -500,6 +500,12 @@ <h2 property="schema:name">Protocol</h2>
                 subscription requirements of the notification protocol.
               </p>
 
+              <p id="json-ld-format">This specification uses JSON-LD [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>] as the preferred data format, and <code>https://www.w3.org/ns/solid/notification/v1</code> as the URI for the JSON-LD Context and as a value of the <code>profile</code> parameter in content negotiation.</p>
+
+              <p id="json-ld-context">JSON-LD representations employing the JSON-LD context MUST ensure that the URI <code>https://www.w3.org/ns/solid/notification/v1</code> is the only value or the terms defined within the context are not overridden by another context.</p>
```
RubenVerborgh, line 505: (just noting that the MUST here is not on a client or server, but on a non-actor)

protocol.html
```diff
@@ -500,6 +500,12 @@ <h2 property="schema:name">Protocol</h2>
...
+
+              <p id="server-json-ld-context-profile-unrecognised">Servers MUST reject HTTP requests with unrecognised JSON-LD context in representation data and values in the <code>profile</code> parameter.</p>
```
RubenVerborgh, line 507:
```suggestion
              <p id="server-json-ld-context-profile-unrecognised">Servers MUST reject HTTP requests with an unrecognised JSON-LD context in representation data and values in the <code>profile</code> parameter.</p>
```

RubenVerborgh (commented):
I can _ok_ this, but I think we're now ad-hoc tackling some of https://github.com/solid/notifications/issues/105; I'd like to see those then documented elsewhere too (future issue/PR is fine).

protocol.html
csarven: Good note.

https://github.com/solid/notifications/pull/120 propose to identify client, resource server, and notification server as the classes of products for conforming implementations. I didn't want to split the statement into three conformance clauses until we had rough consensus.

Alternatively, "content" (of type, meaning, and format as defined in the specification) could be introduced as a product class. The current text is closer to expressing the conformance clause with "content".

Not sure which approach would be preferable - it depends on how we categorise the other requirements. Can we come back to this?

Will look into including JSON-LD concepts/behaviours here - overridden is fine but perhaps should mentioned "protected"?

protocol.html
```diff
@@ -500,6 +500,12 @@ <h2 property="schema:name">Protocol</h2>
                 subscription requirements of the notification protocol.
               </p>
 
+              <p id="json-ld-format">This specification uses JSON-LD [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>] as the preferred data format, and <code>https://www.w3.org/ns/solid/notification/v1</code> as the URI for the JSON-LD Context and as a value of the <code>profile</code> parameter in content negotiation.</p>
```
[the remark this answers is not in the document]
csarven: The intention was to say that when `profile` is used, that's the URI that MUST be used. (In addition to using the same URI for the JSON-LD context when context is used.)

I can clarify the statement (or split into multiple statements) once we have rough consensus.

protocol.html
acoburn: So long as we refer to JSON-LD 1.1 and so long as this context uses the `@protected` parameter in its definition (it does), then the second sentence here is unnecessary.

acoburn (approved):

protocol.html
csarven: Removing `#json-ld-context` is okay with me. @RubenVerborgh ?

However, `#json-ld-format` may need to be updated or another paragraph needs be introduced to express that `#expanded-document-form` (using the vocabulary URIs directly) would also conform. Or would you consider that to be the case and not conflicting with the way `#json-ld-format` is currently expressed?

RubenVerborgh: I am okay with that (and it's good input for security considerations / RDF serializations in general).

elf-pavlik (approved):

protocol.html
csarven: Resolving thread with changes in https://github.com/solid/notifications/pull/122/commits/0d4afa35455909e5195a990f0a2883c0b834e542 , https://github.com/solid/notifications/pull/122/commits/4a0b6be9f1cc889c80ae11b86e67c0f192a20751

Issue #123, opened 2022-11-22
Use activitystreams-vocabulary instead of activitystreams-core for notification-data-model


csarven:
The `#notification-data-model` need to be only concerned about the AS2 vocabulary.

PR #124, opened 2022-11-24
Update notification-data-model

csarven:
This PR further describes the Notification Data Model. It is an alternative to PR https://github.com/solid/notifications/pull/60 .

I decided to create this PR taking in the reviews/considerations in PR 60 as well as aim to resolving https://github.com/solid/notifications/issues/123 . Felt it was simpler to create this PR then to (re)work 60 at this point.

The two examples show typical cases: "update" and "add" (append, e.g., to a container/collection). We can indeed add examples for "create", "delete", or other activities through this PR if so desired - I just wanted to get the basics out. We don't have to exemplify every activity here since we are referring to the AS vocabulary any way. Just need to be sufficiently clear about how that ties to the Notifications Protocol.

I want to point again the considerations in using topic/object: https://github.com/solid/notifications/issues/62#issuecomment-1315944696

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -716,28 +716,56 @@ content: "";
           <section id="notification-data-model" inlist="" rel="schema:hasPart" resource="#notification-data-model">
             <h2 property="schema:name">Notification Data Model</h2>
             <div datatype="rdf:HTML" property="schema:description">
-              <p>The content of a Solid Notification is defined in terms of a <cite><a href="https://www.w3.org/TR/activitystreams-core/">Activity Streams 2.0</a></cite> [<cite><a class="bibref" href="#bib-activitystreams-core">ACTIVITYSTREAMS-CORE</a></cite>]. As such, JSON-LD 1.0 is a required baseline serialization format, though other formats are not prohibited [<cite><a class="bibref" href="#bib-json-ld">JSON-LD</a></cite>]. JSON-LD version 1.1 is recommended when using JSON-LD serializations [<cite><a class="bibref" href="#bib-json-ld">JSON-LD11</a></cite>].</p>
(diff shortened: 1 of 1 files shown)


protocol.html
```diff
@@ -716,28 +716,56 @@ <h2 property="schema:name">Notification Features</h2>
...
+<code>  ],</code>
+<code>  "id": "urn:uuid:a4da6738-6be0-11ed-90bd-1be4ba6b6b33",</code>
+<code>  "type": "Add",</code>
+<code>  "object": "https://example.org/guinan/inbox/notification",</code>
+<code>  "target": "https://example.org/guinan/inbox/",</code>
+<code>  "published": "2022-11-24T11:55:31.483Z"</code>
 <code>}</code></pre>
 
-              <p>Implementations can augment the JSON-LD context definition and extend the content of the notifications.</p>
+                <figcaption property="schema:name">An activity indicating that an object (notification) was added to the target (inbox).</figcaption>
```
elf-pavlik, line 767: I wouldn't use example with notification about notification. While they could be used this way I think it adds possibility of confusion without adding any value. @csarven could you change the example to adding anything else than a notification to a container? A calendar event, a photo to the album, a card to the address book, a song to a playlist, anything will do here.

elf-pavlik (changes requested):

protocol.html
csarven: Okay with me https://github.com/solid/notifications/pull/124/commits/c9a5885a206c1597a0f2eef1f2660b78e1c8f016 . Thanks.

elf-pavlik (approved):
LGTM

acoburn (approved):

PR #128, opened 2022-12-05
adjust terminology to match product classes

elf-pavlik:
Depends on #120 

Only uses *subscription client* in two definitions, following the product class naming.


csarven:
Merging as part of https://github.com/solid/notifications/pull/120 based on agreement in https://github.com/solid/notifications-panel/blob/main/meetings/2022-12-01.md#update-conformance-model

Issue #111, opened 2022-09-21
Clarify subscription start and end times


csarven:
As per #feature-expiration , clients can use `notify:expiration` in the #subscription-request to propose the expiration datetime of a subscription.

It'd help to clarify when the client wants a subscription to start.

Thus it may be preferable to favour the notions "start" and "end" datetimes, instead of "expiration" - meanwhile acknowledging that "expiration" is semantically equivalent to the "end" datetime.

The #subscription-response should indicate start and end datetimes that the server is committing to at the time of the message.

csarven:
Clients can omit the start datetime in the subscription-request to indicate the subscription can start "as soon as possible".

Clients can omit the end datetime in the subscription-request to indicate the subscription can continue for unspecified period of time.

(There may be more useful interpretations for missing start and/or endTimes.)

elf-pavlik:
I think it might be simpler to leave the *expiration* completely up to the notifications publisher and always provide a way for the subscriber to *unsubscribe*. I don't think we need to renew the subscription since a new one can be created instead.

What are the specific use cases which we are trying to solve with expiration/start/end?

I only can think of one that the client might want to end the subscription sooner than the server decided to expire it. And here unsubscribing seems simpler to me.

csarven:
Again, it is optional. It is to clarify the model for the subscription period, which is an event in and itself, with an actual start/end times (including now and indefinitely). That information can be declarative (and recorded) or imperative (derived) from other values or activity elsewhere in the system. It most certainly plays a role in data auditing at least for the server, and it is certainly a "receipt" for the client. Client doesn't have to provide the start time but the server certainly knows when a subscription is active or valid from. Client can certainly provide the start time when it is actually capable of receiving updates. There is absolutely no need to couple the action to request a subscription with the desired datetime to receive updates. It is not always "now". Keep different subscription types in mind.

The current features, `expiration` and `rate,` are essentially trying to capture the rules around this. There are age old standards that describe events with start/end times, periodicity, recurring rules, among other things. See RFC5545 (`DTSTART`/`DTEND`), schema.org's Event (`schema:startDate`/`schema:endDate`) or ActivityStreams2 (`as:startTime`/`as:endTime`) for instance. Ditto "duration"...

I'm essentially proposing that we re-use terminology and modelling from existing stuff.

elf-pavlik:
[...]

PR #135, opened 2022-12-20
Update data models

csarven:
Follows consensus of https://github.com/solid/notifications-panel/blob/main/meetings/2022-12-15.md

Resolves https://github.com/solid/notifications/issues/71 https://github.com/solid/notifications/issues/111 .

---

[Preview](http://htmlpreview.github.io/?https://github.com/solid/notifications/blob/6f4440357bee5a2f4f53bec3d6e844c637f669f1/protocol.html) | [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fnotifications%2Fc3bff4258be16d906f061ea89005de6c43abeb98%2Fprotocol.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fnotifications%2F6f4440357bee5a2f4f53bec3d6e844c637f669f1%2Fprotocol.html)

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -303,9 +303,8 @@ content: "";
                   <a class="tocxref" href="#introduction"><span class="secno">1</span> <span class="content">Introduction</span></a>
                   <ol>
                     <li><a href="#overview"><span class="secno">1.1</span> <span class="content">Overview</span></a></li>
-                    <li><a href="#terminology"><span class="secno">1.2</span> <span class="content">Terminology</span></a></li>
-                    <li><a href="#namespaces"><span class="secno">1.3</span> <span class="content">Namespaces</span></a></li>
-                    <li><a href="#conformance"><span class="secno">1.4</span> <span class="content">Conformance</span></a></li>
+                    <li><a href="#namespaces"><span class="secno">1.2</span> <span class="content">Namespaces</span></a></li>
+                    <li><a href="#conformance"><span class="secno">1.3</span> <span class="content">Conformance</span></a></li>
                   </ol>
                 </li>
                 <li class="tocline">
@@ -317,25 +316,37 @@ content: "";
                     <li class="tocline">
                       <a class="tocxref" href="#subscription"><span class="secno">2.2</span> Subscription</a>
                     </li>
+                    <li class="tocline">
+                      <a class="tocxref" href="#notification-channel"><span class="secno">2.3</span> <span class="content">Notification Channel</span></a>
+                    </li>
+                    <li class="tocline">
+                      <a class="tocxref" href="#notification-message"><span class="secno">2.4</span> <span class="content">Notification Message</span></a>
+                    </li>
+                  </ol>
+                </li>
+
+                <li class="tocline">
+                  <a class="tocxref" href="#data-model"><span class="secno">3</span> <span class="content">Data Model</span></a>
+                  <ol>
+                    <li class="tocline">
+                      <a class="tocxref" href="#description-resource-data-model"><span class="secno">3.1</span> <span class="content">Description Resource</span></a>
+                    </li>
+                    <li class="tocline">
+                      <a class="tocxref" href="#subscription-resource-data-model"><span class="secno">3.2</span> <span class="content">Subscription Resource</span></a>
+                    </li>
+                    <li class="tocline">
+                      <a class="tocxref" href="#notification-channel-data-model"><span class="secno">3.3</span> <span class="content">Notification Channel</span></a>
+                    </li>
+                    <li class="tocline">
+                      <a class="tocxref" href="#notification-message-data-model"><span class="secno">3.4</span> <span class="content">Notification Message</span></a>
+                    </li>
                   </ol>
                 </li>
                 <li class="tocline">
-                  <a class="tocxref" href="#authentication-authorization"><span class="secno">3</span> <span class="content">Authentication and Authorization</span></a>
-                </li>
-                <li class="tocline">
-                  <a class="tocxref" href="#notification-channel-types"><span class="secno">4</span> <span class="content">Notification Channel Types</span></a>
-                </li>
-                <li class="tocline">
-                  <a class="tocxref" href="#notification-features"><span class="secno">5</span> <span class="content">Notification Features</span></a>
-                </li>
-                <li class="tocline">
-                  <a class="tocxref" href="#notification-data-model"><span class="secno">6</span> <span class="content">Notification Data Model</span></a>
-                </li>
-                <li class="tocline">
-                  <a class="tocxref" href="#extensions"><span class="secno">7</span> <span class="content">Extensions</span></a>
+                  <a class="tocxref" href="#extensions"><span class="secno">4</span> <span class="content">Extensions</span></a>
                   <ol>
-                    <li><a href="#notification-channel-types-extension"><span class="secno">7.1</span> <span class="content">Notification Channel Types Extension</span></a></li>
-                    <li><a href="#notification-features-extension"><span class="secno">7.2</span> <span class="content">Notification Features Extension</span></a></li>
+                    <li><a href="#notification-channel-types-extension"><span class="secno">4.1</span> <span class="content">Notification Channel Types</span></a></li>
+                    <li><a href="#notification-features-extension"><span class="secno">4.2</span> <span class="content">Notification Features</span></a></li>
                   </ol>
                 </li>
                 <li class="tocline">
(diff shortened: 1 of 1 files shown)


csarven:
https://github.com/solid/notifications/pull/135/commits/b05cc76d3343321799c79322f4a95d28cbf12290 is based on consensus in  2022-12-20 minutes: https://github.com/solid/notifications-panel/pull/86

csarven:
`startTime` and `endTime` are features and originally intended to be part of the notifications vocab and JSON-LD context. They are also however defined in AS2. Do we need to do alias magic for the context? How would we work it out in the notify vocab?

elf-pavlik (approved):
I would consider reconciling terminology section and data model sections. It looks like each term has it's own section in the document. It might make sense to just move the definition from terminology to that section and remove *Data Model* suffix. This way concepts are defined and specified in the same place, it should also simplify linking to those concepts.

protocol.html
```diff
@@ -563,61 +563,144 @@ <h3 property="schema:name"><span>Issue</span>: Solid Notification Vocabulary and
               <section id="discovery" inlist="" rel="schema:hasPart" resource="#discovery">
                 <h3 property="schema:name">Discovery</h3>
                 <div datatype="rdf:HTML" property="schema:description">
-                  <section id="notification-channel-discovery" inlist="" rel="schema:hasPart" resource="#notification-channel-discovery">
-                    <h4 property="schema:name">Notification Channel Discovery</h4>
+                  <p>This section describes the discovery of <cite><a href="#description-resource-discovery" rel="rdfs:seeAlso">Description Resource</a></cite> that includes the <cite><a href="#subscription-resource-data-model" rel="rdfs:seeAlso">Subscription Resource Data Model</a></cite> and <cite><a href="#notification-channel-data-model" rel="rdfs:seeAlso">Notification Channel Data Model</a></cite>.</p>
+
+                  <section id="description-resource-discovery" inlist="" rel="schema:hasPart" resource="#description-resource-discovery">
+                    <h4 property="schema:name">Description Resource Discovery</h4>
```
elf-pavlik, line 569:
```suggestion
```

I don't think this sub section is needed. given line 566 above already stating:

> This section describes the discovery of [Description Resource](https://htmlpreview.github.io/?https://github.com/solid/notifications/blob/b05cc76d3343321799c79322f4a95d28cbf12290/protocol.html#description-resource-discovery) that includes the [Subscription Resource Data Model](https://htmlpreview.github.io/?https://github.com/solid/notifications/blob/b05cc76d3343321799c79322f4a95d28cbf12290/protocol.html#subscription-resource-data-model) and [Notification Channel Data Model](https://htmlpreview.github.io/?https://github.com/solid/notifications/blob/b05cc76d3343321799c79322f4a95d28cbf12290/protocol.html#notification-channel-data-model).

protocol.html
```diff
@@ -563,61 +563,144 @@ <h3 property="schema:name"><span>Issue</span>: Solid Notification Vocabulary and
...
+                            <a rel="spec:requirementSubject" href="#ResourceServer">Resource Servers</a>
+                            <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> accept requests targeting the <a href="#description-resource">description resource</a> when the value of the <code>Accept</code> header indicates a preferred representation in <code>application/ld+json</code> [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>].
                           </span>
                         </span>
                       </p>
 
-                      <p>The data model of the notification channel and its capabilities is defined in <cite><a href="#notification-channel-conformance" rel="rdfs:seeAlso">Notification Channel Conformance</a></cite>.</p>
+                      <p>The <a href="#description-resource">description resource</a> has the following properties:</p>
```
elf-pavlik, line 586:
```suggestion
   <h4 property="schema:name">Description Resource Data Model</h4>
                      <p>The <a href="#description-resource">description resource</a> has the following properties:</p>
```

Or without *Data Model* suffix if we reconcile it with the term definition.
This change would also require a `<section>` with `id` around it

protocol.html
```diff
@@ -563,61 +563,144 @@ <h3 property="schema:name"><span>Issue</span>: Solid Notification Vocabulary and
...
-                    <h4 property="schema:name">Notification Channel Conformance</h4>
+                  <section id="notification-channel-data-model" inlist="" rel="schema:hasPart" resource="#notification-channel-data-model">
+                    <h4 property="schema:name">Notification Channel Data Model</h4>
                     <div datatype="rdf:HTML" property="schema:description">
+                      <p>The <a href="#notification-channel">notification channel</a> has the following properties:</p>
 
-                      <p>A Notification Channel has the following properties:</p>
+                      <ul>
+                        <li id="notify-channel-id">One <code>id</code> property to identify the notification channel.</li>
+                        <li id="notify-type">One <code>type</code> property to state the notification channel type (<cite><a href="#notification-channel" rel="rdfs:seeAlso">Notification Channel</a></cite>).</li>
```
elf-pavlik, line 669:
```suggestion
                        <li id="notify-type">One <code>type</code> property to state the notification channel type (<cite><a href="#notification-channel-types" rel="rdfs:seeAlso">Notification Channel Types</a></cite>).</li>
```

elf-pavlik:
We could add non-normative text explaining shortly when subscription can be used and when available channels can be advertised in description resource. (this could also be tracked as separate issue)

Advertising subscription resources can be used to:

* create notification channel for protected resource, which relies on capability url to protect the channel
* create notification channel with specific features required by the subscribing client

Advertising existing channels can be used to:

* provide notification chennel for protected resource, which does NOT rely on capability url to protect the channel
* provide notification channel for public read resource, which does NOT need to by protected

CxRes:
> I would consider reconciling terminology section and data model sections. It looks like each term has it's own section in the document. It might make sense to just move the definition from terminology to that section and remove _Data Model_ suffix. This way concepts are defined and specified in the same place, it should also simplify linking to those concepts.

I am not sure, I follow. Also, you need definitions ideally at the top of the document.

protocol.html
```diff
@@ -699,29 +765,29 @@ <h2 property="schema:name">Authentication and Authorization</h2>
             </div>
           </section>
 
-          <section id="subscription-types" inlist="" rel="schema:hasPart" resource="#subscription-types">
-            <h2 property="schema:name">Subscription Types</h2>
+          <section id="notification-channel-types" inlist="" rel="schema:hasPart" resource="#notification-channel-types">
+            <h2 property="schema:name">Notification Channel Types</h2>
             <div datatype="rdf:HTML" property="schema:description">
-              <p>Subscription types define particular interaction patterns for agents to establish and make use of subscription to resources.</p>
+              <p>Notification channel types define particular interaction patterns for clients to establish and make use of notification channels.</p>
```
CxRes, line 771: Channel types define a communication protocol more than an interaction pattern.

protocol.html
```diff
@@ -699,29 +765,29 @@ <h2 property="schema:name">Authentication and Authorization</h2>
...
             </div>
           </section>
 
           <section id="notification-features" inlist="" rel="schema:hasPart" resource="#notification-features" typeof="skos:ConceptScheme">
             <h2 property="schema:name">Notification Features</h2>
             <div datatype="rdf:HTML" property="schema:description">
-              <p property="skos:definition">Features allow subscription clients to customize the details of a subscription request. Some features are specific to particular subscription types while others may be used across all types. These shared features are listed as an initial baseline, though not all subscription types are required to implement these. Individual subscription types may define type-specific features.
+              <p property="skos:definition">Features allow subscription clients to customize the details of a subscription request. Some features are specific to particular notification channel types while others may be used across all types. These shared features are listed as an initial baseline, though not all notification channel types are required to implement these. Individual notification channel types may define type-specific features.
```
CxRes, line 790: "a particular notification channel type" (singular)

protocol.html
```diff
@@ -730,13 +796,13 @@ <h2 property="schema:name">Notification Features</h2>
...
                 <dt about="#feature-rate" property="skos:prefLabel" typeof="skos:Concept"><dfn id="feature-rate">rate</dfn></dt>
-                <dd about="#feature-rate" property="skos:definition">An ISO 8601 duration value indicating the minimum amount of time to elapse between notifications sent to the client.</dd>
+                <dd about="#feature-rate" property="skos:definition">A <code>xsd:duration</code> value indicating the minimum amount of time to elapse between notifications sent to receiver.</dd>
                 <dt about="#feature-accept" property="skos:prefLabel" typeof="skos:Concept"><dfn id="feature-accept">accept</dfn></dt>
                 <dd about="#feature-accept" property="skos:definition">A MIME-Type value that indicates the desired serialization of the notification. For instance, <code>text/turtle</code> may be acceptable.</dd>
```
CxRes, line 807: Accept should use a definition analogous to HTTP. In this case, content type in the form of MIME type (which is much more to the point than serialization).

acoburn (approved):

csarven:
I'd like to merge this PR after updating it with alternatives to `startTime` and `endTime`:

* `start` and `end`
* `startDateTime` and `endDateTime`
* `startDate` and `endDate`
* `startAt` and `endAt`
* `dtstart` and `dtend`
* `from` and `until`
* `open` and `close`

No doubt that there can potential be a clash. FWIW, `start/endAt` is not used in any vocab that [LOV](https://lov.linkeddata.es/dataset/lov/) has on record. A bit harder to say for JSON-LD contexes. The Solid Notifications vocab and JSON-LD should probably stay clear of AS2, schemaorg, PROV-O, anything in/around Time Ontology, ical .. and maybe some others?

Or maybe even something a bit more complex altogether and use an ObjectProperty that is an instance of an existing of an event type (ical or whatever..).

Or go in the direction of `channelStart` and `channelEnd`, or some other prefix/suffix.

Endless possibilities... but we need to decide in the next five minutes :)

csarven:
Group considered the options and decided to go with `startAt` and `endAt` in 2022-12-27 meeting (PR to minutes follow.)

Feedback still welcome until TR release.

PR #139, opened 2023-01-05
initial StreamingHTTPChannel2023

elf-pavlik:


streaming-http-channel-2023.bs
```diff
@@ -0,0 +1,213 @@
...
+DPoP: <proof>
+Content-Type: application/ld+json
+```
+```jsonld
+{
+  "@context": ["https://www.w3.org/ns/solid/notification/v1"],
+  "type": "StreamingHTTPChannel2023",
+  "topic": "https://storage.example/resource",
+  "state": "opaque-state",
+  "expiration": "2023-12-23T12:37:15Z",
```
csarven, line 133:
```suggestion
  "endAt": "2023-12-23T12:37:15Z",
```

streaming-http-channel-2023.bs
```diff
@@ -0,0 +1,213 @@
...
+  "receiveFrom": "https://fetch.example/aca3cb0a-fb0a-4091-b9f8-8117d2cdb392"
+}
+```
+> Response to the POST request, including channel type and the receiveFrom URL.
+
+In JavaScript, a client can use the data in the response to establish a connection to the Fetch API.
+And define how to handle notifications
+
+```js
+const response = await fetch('https://fetch.example/aca3cb0a-fb0a-4091-b9f8-8117d2cdb392')
```
csarven, line 157: In WebSocket, there is url, protocol parameters where the protocol is the IRI of the channel type. Do we need something along those lines for fetch? Or is the specific receiveFrom URL is intended to be unique for the channel, i.e., channel type need not be part of the request.

streaming-http-channel-2023.bs
```diff
@@ -0,0 +1,213 @@
...
+            "Kjetil Kjernsmo"
+        ],
+        "href": "https://solidproject.org/TR/protocol",
+        "title": "Solid Protocol",
+        "publisher": "W3C Solid Community Group"
+    },
+    "Solid.Notifications": {
+        "authors": [
+            "Aaron Coburn",
+            "Sarven Capadisli"
```
csarven, line 196:
```suggestion
        "editors": [
            "Sarven Capadisli"
        ],                
        "authors": [
            "Aaron Coburn",
            "Sarven Capadisli",
            "elf Pavlik",
            "Rahul Gupta"
```

streaming-http-channel-2023.bs
elf-pavlik: I assume `receiveFrom` to be unique for the channel. Probably this should be made explicit in this spec.

Issue #145, opened 2023-01-12
Define unsubscribing


elf-pavlik:
Currently Notification Protocol only defines Subscription. [WebhookSubscription2021](https://github.com/solid/notifications/blob/main/webhook-subscription-2021.md#5-unsubscribe-api) defines unsubscribing, IMO it should also be defined on the protocol level so that it can be used with any channel type.

I see few aspects that need to be clarified

## Including unsubscribe details in subscription response

This seems reasonable for notification channel description to provide details needed to unsubscribe
It makes sense that authentication should use same identity as subscription client which made subscription request.

## Including unsubscribe details in each notification

WebhookSubscription2021 also requires that unsubscribe details are included in every notification.
Besides that it posts a question in [Further considerations](https://github.com/solid/notifications/blob/main/webhook-subscription-2021.md#further-considerations)

> The fact that the Pod allows any URI to be submitted as a `target` might lead to distributed denial of service attacks originating from Pods. A malicious actor could create an app and trick many users to join it. The malicious app would then create a webhook targeting its desired target. It may be a good idea to build in an automated verification process to confirm that the entity sending the subscription request also owns the target webhook endpoint.

Given above it might make sense to use Capability URL in each notification and don't require authentication on it. This way  one can at least unsubscribe from notifications where malicious actor made the subscription.

/cc @jaxoncreed

CxRes:
Unsubscribe must be handled in the main protocol. I believe it is handled consistently as a subcase of subscription modification, as discussed in #103 and further in meetings.

I propose that we allow clients to POST a partial of Notification Channel Data Model, that is, only the property the clients wants to change to the Subscription Service. In case we wish to unsubscribe, this will simply be an `id` and `endTime`:

```json
{
  "@context": [
    "https://www.w3.org/ns/solid/notifications/v1"
  ],
  "id": "https://channel.example/ac748712",
  "endAt": "time:Instant"
}
```

It can alternatively be a PUT with the above message or even a DELETE on the created subscription endpoint in case the Subscription Service is a container.

This will take away the need to add unsubscribe information in the subscription response or message!

CxRes:
There is one other complication involved, which I had raised in a discussion two years ago https://gitter.im/solid/specification?at=5ea2b8c6501f8f72a5ffa61a. Some clients might want to continue to watch resources that have been deleted, in the expectation that they will be created again, other times we want to stop watching. I believe the proposal I make above is suitable to handle both these cases.

[...]

PR #146, opened 2023-01-12
initial WebhookChannel2023

elf-pavlik:
This is initial version following common template using by other subscription types.
We can add separate primer to provide more detail example like in WebhookSubscription2023.

I also created #145 to define unsubscribing on the protocol level, later webhook can further specify it if needed.


webhook-channel-2023.bs
```diff
@@ -0,0 +1,190 @@
...
+* Notification Sender MUST perform authenticated request to *sendTo* webhook endpoint,
+    using identity provided as *sender* in the subscription response.
+
+As described by the Solid Notifications Protocol section on Authorization,
+the WebhookChannel2023 requires authorization and follows the guidance of the Solid Protocol
+sections on Authentication and Authorization [[!Solid.Protocol]].
+
+It is beyond the scope of this document to describe how a client fetches an access token.
+Solid-OIDC is one example of an authentication mechanism that could be used with Solid Notifications [[!Solid.OIDC]].
+
```
jaxoncreed, line 156: I feel like this section is incomplete.

Pointing to Solid OIDC for the first use case (Subscription client MUST perform authenticated subscription request.) is fine as that use case is very clearly a client-pod relationship which is what Solid OIDC was designed to handle.

There are many questions left unanswered about the second use case (Notification Sender MUST perform authenticated request to *sendTo* webhook endpoint, using identity provided as *sender* in the subscription response.).

- Are we saying that a sender's "authenticated request" needs to be Solid OIDC as implied by the Solid Notifications Protocol?
- If so, what role is the sender? Is the sender the token issuer? Is the sender the token subject?
- In my opinion, Solid OIDC is an inappropriate Authentication protocol for this use case. It's designed for use cases that involve a user agent, and therefore has more complexities than are needed. This use case is a simple server-to-server communications, but the following unnecessary things would be required if we were forced to use Solid OIDC:
  - An identity provider, not for users, but the send itself
  - The use of DPoP tokens which are not needed because DPoP tokens are designed for ecosystems that have three constituents: Identity Provider, Relying Party, and Resource Server. 
 - I've been out of it for a bit. Perhaps Solid OIDC has added the Client Credentials flow. If so, this does solve the DPoP token problem mentioned about as Client Credentials doesn't use DPoP, but it still requires extra complexity because the the sender would need to authenticated with an identity provider before sending notifications.
 - I still prefer the method I outlined in the original document: Just give the sender its own public/private key pair and let it mint a simple token itself without needing to get a Solid OIDC IdP involved. Again, I've been out of it, so if there's an explanation as to why this approach is worse, I'm happy to hear it.
 - But, no matter what approach we go with, I would still prefer to have this section go into a little more detail on how senders send authenticated requests.

jaxoncreed (changes requested):
Thanks for this @elf-pavlik . I'll be sure to hop into the meeting next week and we can talk about things in-person as well.

webhook-channel-2023.bs
```diff
@@ -0,0 +1,190 @@
...
+This specification defines a subscription type that applies these patterns to WebHooks.
+
+## Specification Goals ## {#goals}
+
+In addition to the goals set forth by the [[!Solid.Notifications]] , this specification has goals
+required by the WebHooks use case:
+
+ 1. **Verifiable requests to a notification receiver** - a notification receiver must be able to confirm
+    if a request truly came from a specific notification sender.
+ 2. **Unsubscribing from a WebHook** - Unlike websockets, where sockets can simply be closed by the client,
```
jaxoncreed, line 56: This document doesn't mention how unsubscribing works. Is it described somewhere else?

webhook-channel-2023.bs
elf-pavlik: @jaxoncreed I agreed with all your points, recently I [posted relevant comments on gitter](https://gitter.im/solid/authentication-panel?at=63d81b2b0c94855213d78674)
At the same time I don't think the problem is specific for WebHook channel type. I see it as more common requirement to authenticate server-side clients which we should be able to do in consistent way across the Solid ecosystem.

webhook-channel-2023.bs
elf-pavlik: Below you can see inline issue linking to #145

jaxoncreed:
@elf-pavlik is the idea to just approve this to have something on the books and then fix the problems later?

webhook-channel-2023.bs
elf-pavlik: I'm going to add #155 as inline issue. This way we can already start using Capability URL for *sendTo*, later adding support for Solid-OIDC and other more fitting options.

elf-pavlik:
@jaxoncreed I'd like to add any unresolved issues inline and try to publish 0.1 during next meeting.

webhook-channel-2023.bs
elf-pavlik: image

streaming-http-channel-2023.html
```diff
@@ -0,0 +1,776 @@
...
+    <dd data-md>
+     <p>The <a href="https://solid.github.io/notifications/protocol#notify-receiveFrom">receiveFrom</a> property
+is used in the body of the subscription response.
+The value of source property MUST be a URI, using the <code>https</code> scheme.</p>
+   </dl>
+   <p>A client establishes a subscription using the StreamingHTTPChannel2023 type by sending an authenticated
+subscription request to the Subscription Resource retrieved via Solid Notifications discovery.</p>
+   <p>For StreamingHTTPChannel2023 interactions, the client sends a JSON-LD payload to the appropriate
+Subscription Resource via POST. The only required fields in this interaction are type and topic.
+The type field MUST contain the type of channel being requested: <code>StreamingHTTPChannel2023</code> The topic field MUST contain the URL of the resource a client wishes to subscribe to changes.</p>
```
TallTed, line 647:
```suggestion
   <p>A client establishes a subscription using the <code>StreamingHTTPChannel2023</code> type by sending an authenticated
subscription request to the Subscription Resource retrieved via Solid Notifications discovery.</p>
   <p>For <code>StreamingHTTPChannel2023</code> interactions, the client sends a JSON-LD payload to the appropriate
Subscription Resource via <code>POST</code>. The only required fields in this interaction are <code>type</code> and <code>topic</code>.
The <code>type</code> field MUST contain the type of channel being requested: <code>StreamingHTTPChannel2023</code>. The <code>topic</code> field MUST contain the URL of the resource to which changes a client wishes to subscribe.</p>
```

elf-pavlik: I'm sorry, this file doesn't really belong here and it is already gitignored on `main`. I'm going to adapt this suggestion to the source which is already in the repo

webhook-channel-2023.bs
```diff
@@ -0,0 +1,190 @@
...
+
+**This section is non-normative.**
+
+This specification uses the terms from the [[!Solid.Notifications]] specification.
+
+# WebhookChannel2023 Type # {#channel-type}
+
+This specification defines the WebhookChannel2023 type for use with Solid Notifications channels that use the Webhooks.
+
+An WebhookChannel2023 MUST conform to the [Solid Notifications Protocol](https://solid.github.io/notifications/protocol#discovery).
```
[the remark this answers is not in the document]
TallTed: I concur with @jaxoncreed. I don't think this should be considered resolved until `An WebhookChannel2023` becomes `A WebhookChannel2023`.

webhook-channel-2023.bs
```diff
@@ -0,0 +1,190 @@
...
+
+An WebhookChannel2023 SHOULD support the [Solid Notifications Features](https://solid.github.io/notifications/protocol#notification-features).
```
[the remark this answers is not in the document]
TallTed: I concur with @jaxoncreed. I don't think this should be considered resolved until `An WebhookChannel2023` becomes `A WebhookChannel2023`.

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+* The **topic** field MUST contain the URL of the resource a client wishes to subscribe to changes.
+* The **sendTo** field MUST contain the `https` URL of the webhook endpoint, where the notification sender
+    is expected to send notifications.
+
+For WebhookChannel2023 interactions, the subscription server responds with a subscription response.
+The only required fields in the payload are *id*, *type*, *topic*, *sender*
+* The **id** field MUST be a URI, which identifies created notification channel.
+* The **type** field MUST contain the type of channel created: `WebhookChannel2023` Web
+* The **topic** field MUST contain the URL of the resource which notifications will be about.
+* The **sender** field MUST contain URI, which identifies the notifications sender
```
TallTed, line 99:
```suggestion
* The **`sender`** field MUST contain a URI which identifies the notifications sender.
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+:: The *sendTo* property
+    is used in the body of the subscription request.
+    The value of *sendTo* property MUST be a URI, using the `https` scheme.
+
+A client establishes a subscription using the WebhookChannel2023 type by sending an authenticated
+subscription request to the Subscription Resource retrieved via Solid Notifications discovery.
+
+For WebhookChannel2023 interactions, the subscription client sends a subscription request to an apropriate
+Subscription Resource. The only required fields in the payload are *type*, *topic* and *sendTo*.
+* The **type** field MUST contain the type of channel being requested: `WebhookChannel2023`
```
TallTed, line 89:
```suggestion
* The **`type`** field MUST contain the type of channel being requested: `WebhookChannel2023`.
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+An WebhookChannel2023 MUST conform to the [Solid Notifications Protocol](https://solid.github.io/notifications/protocol#discovery).
+
+An WebhookChannel2023 SHOULD support the [Solid Notifications Features](https://solid.github.io/notifications/protocol#notification-features).
+
+The WebhookChannel2023 type further constrains following properties properties:
+
+: sendTo
+:: The *sendTo* property
+    is used in the body of the subscription request.
+    The value of *sendTo* property MUST be a URI, using the `https` scheme.
```
TallTed, line 82:
```suggestion
: `sendTo`
:: The *`sendTo`* property
    is used in the body of the subscription request.
    The value of the *`sendTo`* property MUST be a URI, using the `https` scheme.
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+
+A client establishes a subscription using the WebhookChannel2023 type by sending an authenticated
+subscription request to the Subscription Resource retrieved via Solid Notifications discovery.
+
+For WebhookChannel2023 interactions, the subscription client sends a subscription request to an apropriate
+Subscription Resource. The only required fields in the payload are *type*, *topic* and *sendTo*.
```
TallTed, line 88:
```suggestion
For WebhookChannel2023 interactions, the subscription client sends a subscription request to an appropriate
Subscription Resource. The only required fields in the payload are *`type`*, *`topic`*, and *`sendTo`*.
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+* The **type** field MUST contain the type of channel being requested: `WebhookChannel2023`
+* The **topic** field MUST contain the URL of the resource a client wishes to subscribe to changes.
+* The **sendTo** field MUST contain the `https` URL of the webhook endpoint, where the notification sender
+    is expected to send notifications.
```
TallTed, line 92:
```suggestion
* The **`topic`** field MUST contain the URL of the resource to which changes a client wishes to subscribe.
* The **`sendTo`** field MUST contain the `https` URL of the webhook endpoint, where the notification sender
    is expected to send notifications.
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+
+For WebhookChannel2023 interactions, the subscription server responds with a subscription response.
+The only required fields in the payload are *id*, *type*, *topic*, *sender*
+* The **id** field MUST be a URI, which identifies created notification channel.
+* The **type** field MUST contain the type of channel created: `WebhookChannel2023` Web
+* The **topic** field MUST contain the URL of the resource which notifications will be about.
```
TallTed, line 98:
```suggestion
For `WebhookChannel2023` interactions, the subscription server responds with a subscription response.
The only required fields in the payload are *`id`*, *`type`*, *`topic`*, *`sender`*
* The **`id`** field MUST be a URI, which identifies the created notification channel.
* The **`type`** field MUST contain the type of channel created: `WebhookChannel2023`.
* The **`topic`** field MUST contain the URL of the resource which notifications will be about.
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+  "@context": ["https://www.w3.org/ns/solid/notification/v1"],
+  "type": "WebhookChannel2023",
+  "topic": "https://storage.example/resource",
+  "sendTo": "https://webhook.example/871b84e7",
+  "state": "opaque-state",
+  "expiration": "2023-12-23T12:37:15Z",
+  "rate": "PT10s"
+}
+```
+> POST request including type, topic and sendTo targeting the Notification Subscription Resource.
```
TallTed, line 127:
```suggestion
> `POST` request including `type`, `topic`, and `sendTo` targeting the Notification Subscription Resource.
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+{
+  "@context": "https://www.w3.org/ns/solid/notification/v1",
+  "id": "https://channels.example/ea1fcf13-7482-4bbb-a6c1-c168ddd7b0aa"
+  "type": "WebhookChannel2023",
+  "sender": "https://sender.example/#it",
+  "expiration": "2023-12-23T12:37:15Z",
+  "rate": "PT10s"
+}
+```
+> Response to the POST request, including channel id, type and the sender identity.
```
TallTed, line 144:
```suggestion
> Response to the `POST` request, including `channel id`, `type`, and `sender identity`.
```

TallTed (changes requested):
Generally minor tweaks. A number of additional instances of `WebhookChannel2023` (and possibly other entities) should also be wrapped in backticks.

jaxoncreed (approved):
Given the fact that issues are listed in this spec, I can approve it.

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+Favicon: https://solidproject.org/TR/solid.svg
+ED: https://solid.github.io/notifications/webhook-channel-2023
+Repository: https://github.com/solid/notifications
+Inline Github Issues: title
+Markup Shorthands: markdown yes
+Max ToC Depth: 2
+Editor: Jackson Morgan
+Editor: [elf Pavlik](https://elf-pavlik.hackers4peace.net/)
+Abstract:
+  The [[!Solid.Notifications]] defines a set of interaction patterns for agents to receive notification
```
csarven, line 18:
```suggestion
  The [[!Solid.Notifications.Protocol]] defines a set of interaction patterns for agents to receive notification
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+
+Issue(155):
+
+Issue(145):
+
+## Terminology ## {#terminology}
+
+**This section is non-normative.**
+
+This specification uses the terms from the [[!Solid.Notifications]] specification.
```
csarven, line 67:
```suggestion
This specification uses the terms from the [[!Solid.Notifications.Protocol]] specification.
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+        ],
+        "href": "https://solidproject.org/TR/protocol",
+        "title": "Solid Protocol",
+        "publisher": "W3C Solid Community Group"
+    },
+    "Solid.Notifications": {
+        "authors": [
+            "Aaron Coburn",
+            "Sarven Capadisli"
+        ],
```
csarven, line 176:
```suggestion
    "Solid.Notifications.Protocol": {
        "editors": [
            "Sarven Capadisli"
        ],                
        "authors": [
            "Aaron Coburn",
            "Sarven Capadisli",
            "elf Pavlik",
            "Rahul Gupta"
        ],
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+  This document was produced by a group operating under the [W3C Community Contributor License Agreement
+  (CLA)](https://www.w3.org/community/about/process/cla/). A human-readable
+  [summary](https://www.w3.org/community/about/process/cla-deed/) is available.
+</pre>
+
+# Introduction # {#introduction}
+
+*This section is non-normative.*
+
+The [[!Solid.Notifications]] describes a general pattern by which agents can be notified when a Solid Resource changes.
```
csarven, line 46:
```suggestion
The [[!Solid.Notifications.Protocol]] describes a general pattern by which agents can be notified when a Solid Resource changes.
```

webhook-channel-2023.bs
```diff
@@ -0,0 +1,192 @@
...
+This specification defines a subscription type that applies these patterns to WebHooks.
+
+## Specification Goals ## {#goals}
+
+In addition to the goals set forth by the [[!Solid.Notifications]] , this specification has goals
```
csarven, line 51:
```suggestion
In addition to the goals set forth by the [[!Solid.Notifications.Protocol]] , this specification has goals
```

PR #149, opened 2023-01-16
Fix Language for v0.2

CxRes:
Made Changes to language for 0.2
+ Fixed use of Subscription in Abstract and Introduction.
+ Added some intro text to Overview to contextualize diagrams.
+ Modified intro text for the Protocol section. Add context for channel types.
+ Cleaned up feature and broke into paragraphs.
+ Fixed Data Models to be more precise.
+ Small semantic and grammar changes.


CxRes:
I have been moderately aggressive with the changes. If you have questions, I would be happy to walk you through why I felt a change was needed.

protocol.html
```diff
@@ -627,7 +632,12 @@ <h5 property="schema:name"><span>Note</span>: Notification Channel Type Registry
...
+                      <p property="skos:definition">Features describe custom properties of a notification channel.
+
+                      <p>Discription Resources may advertize a list of features that the <a href="#SubscriptionServer">Subscription Server</a> can modify.
+                      <a href="#SubscriptionClient">Subscription Clients</a> can modify these features to customise the details of a requested notification channel. Subscription clients can discover available features in a <a href="#subscription-resource">subscription resource</a>.</p>
```
csarven, line 638:
```suggestion
                      <p property="skos:definition">Features describe custom properties of a notification channel. Description Resources may advertise a list of features that the <a href="#SubscriptionServer">Subscription Server</a> can modify. <a href="#SubscriptionClient">Subscription Clients</a> can modify these features to customise the details of a requested notification channel. Subscription clients can discover available features in a <a href="#subscription-resource">subscription resource</a>.</p>
```

protocol.html
```diff
@@ -609,14 +614,14 @@ <h3 property="schema:name">Notification Channel</h3>
                   <section id="notification-channel-types" inlist="" rel="schema:hasPart" resource="#notification-channel-types">
                     <h4 property="schema:name">Notification Channel Types</h4>
                     <div datatype="rdf:HTML" property="schema:description">
-                      <p>A notification channel type is a specific communication protocol that governs how classes of products can establish a connection to send and receive messages. A notification channel type also specifies the data model and format of the messages that are exchanged between products. A notification channel type is identified by a IRI for use with both <a href="#discovery">discovery</a> and the establishment of notification channels.</p>
+                      <p>A notification channel type is a specific communication protocol that governs how classes of products can establish a connection to send and receive messages. A notification channel type is identified by a IRI used to both <a href="#discovery">discover</a> and establish the <a href="#notification-channel">notification channels</a>.</p>
```
csarven, line 617: I think the following information is significant enough to be mentioned here:

>specifies the data model and format of the messages that are exchanged between products

```suggestion
                      <p>A notification channel type is a specific communication protocol that governs how classes of products can establish a connection to send and receive messages. A notification channel type is identified by a IRI used to both <a href="#discovery">discover</a> and establish the <a href="#notification-channel">notification channels</a>. A notification channel type also specifies the data model and format of the messages that are exchanged between products.</p>
```

protocol.html
```diff
@@ -527,12 +530,14 @@ <h3 property="schema:name">Discovery</h3>
...
 
                   <p about="" id="resource-server-description-resource-accept" rel="spec:requirement" resource="#resource-server-description-resource-accept"><span property="spec:statement"><a rel="spec:requirementSubject" href="#ResourceServer">Resource Servers</a> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> accept requests targeting the <a href="#description-resource">description resource</a> when the value of the <code>Accept</code> header indicates a preferred representation in <code>application/ld+json</code> [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>].</span></p>
 
+                  <p><span property="spec:statement"><a rel="spec:requirementSubject" href="#ResourceServer">Resource Servers</a> <span rel="spec:requirementLevel" resource="spec:MAY">MAY</span> optionally accept requests targeting the <a href="#description-resource">description resource</a> when the value of the <code>Accept</code> header indicates preferred representation other than <code>application/ld+json</code> [<cite><a class="bibref" href="#bib-json-ld11">JSON-LD11</a></cite>].</p>
+
```
csarven, line 540: This need not be said because the option is there by design (RFC7231). Many things are optional. Mostly need to focus on minimal requirements towards interop.

```suggestion
```

protocol.html
```diff
@@ -527,12 +530,14 @@ <h3 property="schema:name">Discovery</h3>
...
+                  <p about="" id="resource-server-link-resource-description-resource" rel="spec:requirement" resource="#resource-server-link-resource-description-resource"><span property="spec:statement">A <a href="#ResourceServer" rel="spec:requirementSubject">Resource Server</a> that wants to enable <a href="#SubscriptionClient">Subscription Clients</a> to discover subscription resources and notification channels available for a given resource <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> advertise the associated resources describing the <a href="#subscription-resource">subscription resources</a> and <a href="#notification-channel">notification channels</a> by responding to an HTTP request including a <code>Link</code> header with the <code>rel</code> value of <code>describedby</code> [<cite><a class="bibref" href="#bib-powder-dr">POWDER-DR</a></cite>] and the Description Resource as link target [<cite><a class="bibref" href="#bib-rfc8288">RFC8288</a></cite>].</span></p>
```
csarven, line 533:
```suggestion
                  <p about="" id="resource-server-link-resource-description-resource" rel="spec:requirement" resource="#resource-server-link-resource-description-resource"><span property="spec:statement"><a href="#ResourceServer" rel="spec:requirementSubject">Resource Servers</a> that want to enable <a href="#SubscriptionClient">Subscription Clients</a> to discover subscription resources and notification channels available for a given resource <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> advertise the associated resources describing the <a href="#subscription-resource">subscription resources</a> and <a href="#notification-channel">notification channels</a> by responding to an HTTP request including a <code>Link</code> header with the <code>rel</code> value of <code>describedby</code> [<cite><a class="bibref" href="#bib-powder-dr">POWDER-DR</a></cite>] and the Description Resource as link target [<cite><a class="bibref" href="#bib-rfc8288">RFC8288</a></cite>].</span></p>
```

protocol.html
```diff
@@ -527,12 +530,14 @@ <h3 property="schema:name">Discovery</h3>
...
...    <p about="" id="resource-server-link-storage-description-resource" rel="spec:requirement" resource="#resource-server-link-storage-description-resource"><span property="spec:statement">When a resource server wants to enable <a href="#SubscriptionClient">Subscription Clients</a> to discover subscription resources and notification channels available to a <a href="https://solidproject.org/TR/protocol#storage">storage</a> in which a given <a href="https://solidproject.org/TR/protocol#resource">resource</a> is in, the <a rel="spec:requirementSubject" href="#ResourceServer">Resource Server</a> <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> advertise the associated resources describing the <a href="#subscription-resource">subscription resources</a> and <a href="#notification-channel">notification channels</a> by responding to an HTTP request including a <code>Link</code> header with the <code>rel</code> value of <code>http://www.w3.org/ns/solid#storageDescription</code> [<cite><a class="bibref" href="#bib-solid-protocol">SOLID-PROTOCOL</a></cite>] and the Description Resource as link target [<cite><a class="bibref" href="#bib-rfc8288">RFC8288</a></cite>].</span></p>
```
csarven, line 535:
```suggestion
                  <p about="" id="resource-server-link-storage-description-resource" rel="spec:requirement" resource="#resource-server-link-storage-description-resource"><span property="spec:statement"><a href="#ResourceServer" rel="spec:requirementSubject">Resource Servers</a> that want to enable <a href="#SubscriptionClient">Subscription Clients</a> to discover subscription resources and notification channels available to a <a href="https://solidproject.org/TR/protocol#storage">storage</a> in which a given <a href="https://solidproject.org/TR/protocol#resource">resource</a> is in <span rel="spec:requirementLevel" resource="spec:MUST">MUST</span> advertise the associated resources describing the <a href="#subscription-resource">subscription resources</a> and <a href="#notification-channel">notification channels</a> by responding to an HTTP request including a <code>Link</code> header with the <code>rel</code> value of <code>http://www.w3.org/ns/solid#storageDescription</code> [<cite><a class="bibref" href="#bib-solid-protocol">SOLID-PROTOCOL</a></cite>] and the Description Resource as link target [<cite><a class="bibref" href="#bib-rfc8288">RFC8288</a></cite>].</span></p>
```

protocol.html
```diff
@@ -494,31 +496,32 @@ <h4 property="schema:name skos:prefLabel">Interoperability</h4>
...
-                  <p>Advertising <a href="#subscription-resource">subscription resources</a> can be used to create notification channels:</p>
+                  <p>Resource Servers may advertize <a href="#subscription-resource">subscription resources</a> that can be used to create notification channels:</p>
 
                   <ul>
                     <li>for protected resource, which relies on capability URL to protect the channel;</li>
                     <li>with specific features required by the subscribing client.</li>
                   </ul>
 
-                  <p>Advertising existing <a href="#notification-channel">notification channels</a> can be used to provide notification channels:</p>
+                  <p>Resource Servers may advertize existing <a href="#notification-channel">notification channels</a> to provide notifications:</p>
```
csarven, line 524: Prefer to stick to advisement-levels language and stay clear from requirement-levels language.

```suggestion
                  <p>Resource Servers can advertise existing <a href="#notification-channel">notification channels</a> to provide notifications:</p>
```

protocol.html
```diff
@@ -494,31 +496,32 @@ <h4 property="schema:name skos:prefLabel">Interoperability</h4>
...
+                  <p><a href="#ResourceServer">Resource Servers</a> enable <a href="#SubscriptionClient">Subscription Clients</a> to discover <a href="#description-resource">description resources</a> that are associated with a topic resource.</p>
 
                   <p about="#description-resource" property="skos:definition" typeof="skos:Concept">A <strong><dfn id="description-resource" property="skos:prefLabel">description resource</dfn></strong> is an <em>RDF document</em> [<cite><a class="bibref" href="#bib-rdf11-concepts">RDF11-CONCEPTS</a></cite>] that includes information about the capabilities and features of a <cite><a href="#subscription-resource-data-model" rel="rdfs:seeAlso">subscription resource</a></cite> or <cite><a href="#notification-channel-data-model" rel="rdfs:seeAlso">notification channel</a></cite>.</p>
 
-                  <p>Advertising <a href="#subscription-resource">subscription resources</a> can be used to create notification channels:</p>
+                  <p>Resource Servers may advertize <a href="#subscription-resource">subscription resources</a> that can be used to create notification channels:</p>
```
csarven, line 517: Prefer to stick to advisement-levels language and stay clear from requirement-levels language.

```suggestion
                  <p>Resource Servers can advertise <a href="#subscription-resource">subscription resources</a> that can be used to create notification channels:</p>
```

protocol.html
```diff
@@ -494,31 +496,32 @@ <h4 property="schema:name skos:prefLabel">Interoperability</h4>
...
                 <h3 property="schema:name">Discovery</h3>
                 <div datatype="rdf:HTML" property="schema:description">
-                  <p about="#topic-resource" property="skos:definition" typeof="skos:Concept">The starting point for discovery is the resource which the notification is about: the <strong><dfn id="topic-resource" property="skos:prefLabel">topic resource</dfn></strong>. Choosing the most appropriate topic resource from which to begin discovery is at the discretion of the <a href="#SubscriptionClient">Subscription Client</a>, since notifications may be available for any resource for authorized access subjects.</p>
+                  <p about="#topic-resource" property="skos:definition" typeof="skos:Concept">The starting point for discovery is the resource which the notification is sought; we call this a <strong><dfn id="topic-resource" property="skos:prefLabel">topic resource</dfn></strong>. Choosing the most appropriate topic resource from which to begin discovery is at the discretion of the <a href="#SubscriptionClient">Subscription Client</a>, since notifications may be available for any resource for authorized access subjects.</p>
```
csarven, line 511: Don't like "we" in a spec!

```suggestion
                  <p about="#topic-resource" property="skos:definition" typeof="skos:Concept">The starting point for discovery is the resource which the notification is about, hereafter referred as the <strong><dfn id="topic-resource" property="skos:prefLabel">topic resource</dfn></strong>. Choosing the most appropriate topic resource from which to begin discovery is at the discretion of the <a href="#SubscriptionClient">Subscription Client</a>, since notifications may be available for any resource for authorized access subjects.</p>
```

protocol.html
```diff
@@ -376,24 +375,27 @@ <h2 about="#introduction" property="schema:name" typeof="deo:Introduction">Intro
...
+                  <p>The following diagram illustrates the flow of data for those notification channel types in which a client establishes a subscription and then maintains a persistent connection to a notifications source are illustrated by the following diagram.</p>
 
                   <figure id="solid-notifications-flow-receivefrom" rel="schema:hasPart" resource="#solid-notifications-flow-receivefrom">
                     <img rel="schema:image" src="notifications-flow-receivefrom.mmd.svg" width="800" />
 
                     <figcaption property="schema:name">Solid Notifications Flow Receive From, where the receiver establishes connection to sender.</figcaption>
                   </figure>
 
-                  <p>Notification channel types in which a client establishes a subscription and does not maintain a persistent connection to a notifications source are illustrated by the following diagram.</p>
+                  <p>The following diagram illustrates  the flow of data for those Notification channel types in which a client establishes a subscription and does not maintain a persistent connection to a notifications source:</p>
```
csarven, line 398:
```suggestion
                  <p>The following diagram illustrates the flow of data for those Notification channel types in which a client establishes a subscription and does not maintain a persistent connection to a notifications source:</p>
```

protocol.html
```diff
@@ -376,24 +375,27 @@ <h2 about="#introduction" property="schema:name" typeof="deo:Introduction">Intro
...
 
                   <figure id="solid-notifications-flow-discovery-subscription" rel="schema:hasPart" resource="#solid-notifications-flow-discovery-subscription">
                     <img rel="schema:image" src="notifications-flow-discovery-subscription.mmd.svg" width="800" />
 
                     <figcaption property="schema:name">Solid Notifications Flow Discovery and Subscription</figcaption>
                   </figure>
 
-                  <p>Notification channel types in which a client establishes a subscription and then maintains a persistent connection to a notifications source are illustrated by the following diagram.</p>
+
+                  <p>A notification channel allows Notification Senders to send notifications to a Notifications Recievers. A notification channel builds upon one of many established web protocols to exchange data, determining its <em>type</em>.</p>
```
csarven, line 388:
```suggestion
                  <p>A notification channel allows a Notification Sender to send notifications to a Notifications Receiver. A notification channel builds upon one of many established web protocols to exchange data, determining its <em>type</em>.</p>
```

protocol.html
```diff
@@ -376,24 +375,27 @@ <h2 about="#introduction" property="schema:name" typeof="deo:Introduction">Intro
                 <h3 property="schema:name">Overview</h3>
                 <div datatype="rdf:HTML" property="schema:description">
                   <p><em>This section is non-normative.</em></p>
-
-                  <p>The following diagram shows the high-level interactions involved discovering notification channels and subscription resources to towards receiving notifications.</p>
+                  <p>The Solid Notification protocol defines a mechanism for client applications to discover notification services avaiable for any given resource.</p>
+                  <p>The following diagram shows the high-level interactions involved in discovering notification channels and subscription resources. A subscription resource allow Subscription Clients to request a customized notification channels.</p>
```
csarven, line 379:
```suggestion
                  <p>The following diagram shows the high-level interactions involved in discovering notification channels and subscription resources. A subscription resource allow Subscription Clients to request a customised notification channels.</p>
```

csarven (changes requested):
Thank you for going through this! Mostly looks good. I do have some objections to changes, made some suggestions.

I don't know how to "undo" some of the changes you've proposed. So, some of the suggestions are essentially updating your changes with the original.

protocol.html
```diff
@@ -906,10 +916,10 @@ <h3 property="schema:name">Security and Privacy Review</h3>
...
-                    <dd about="#security-privacy-review-first-third-party" datatype="rdf:HTML" property="schema:description">Inapplicable.</dd>
+                    <dd about="#security-privacy-review-first-third-party" datatype="rdf:HTML" property="schema:description">Not Applicable.</dd>
 
                     <dt about="#security-privacy-review-private-browsing" id="security-privacy-review-private-browsing"><a href="https://www.w3.org/TR/security-privacy-questionnaire/#private-browsing" rel="cito:repliesTo">How do the features in this specification work in the context of a browser’s Private Browsing or Incognito mode?</a></dt>
-                    <dd about="#security-privacy-review-private-browsing" datatype="rdf:HTML" property="schema:description">No different than <q>browser’s 'normal' state</q>.</dd>
+                    <dd about="#security-privacy-review-private-browsing" datatype="rdf:HTML" property="schema:description">No different than in the <q>browser’s 'normal' mode</q>.</dd>
```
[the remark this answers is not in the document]
CxRes: The reason I had proposed this change was because the question uses "mode". Also browser vendors tend to call it "mode" as well. Can you clarify where you have picked this quote from?

protocol.html
```diff
@@ -738,21 +748,21 @@ <h3 property="schema:name">Notification Channel</h3>
                   <ul>
                     <li id="notify-channel-id">One <code>id</code> property to identify the notification channel.</li>
                     <li id="notify-channel-type">One <code>type</code> property to state the <a href="#notification-channel-types" rel="rdfs:seeAlso">notification channel type</a>.</li>
-                    <li id="notify-topic">At least one <code rel="dcterms:subject" resource="http://www.w3.org/ns/solid/notifications#topic">topic</code> property to identify the resources that the notifications are about.</li>
-                    <li id="notify-feature-types">Zero, one or many property values stating a particular <a href="#notification-features" rel="rdfs:seeAlso">feature</a>.</li>
+                    <li id="notify-topic">A <code rel="dcterms:subject" resource="http://www.w3.org/ns/solid/notifications#topic">topic</code> property with the property value containing at least entry to identify the (<cite><a href="#topic-resource">topic</a></cite>) resources that the notifications are about.</li>
```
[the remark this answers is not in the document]
CxRes: I am accepting this reversal to be consistent with other data model reversals. See my comment below about data model changes.

protocol.html
CxRes: Can you clarify if we are using British or American English? With British English your suggestion here is correct. However, W3C Style Guide expects American English. In which case we should use "customized".

protocol.html
CxRes: I understand what you are saying! But if something like this can be included even non-normatively, it is a good nudge to server developers, makes client life easier.

Irrespective, I shall accept the reversal after you have read/commented.

protocol.html
```diff
@@ -671,8 +681,8 @@ <h3 property="schema:name">Description Resource</h3>
 
                   <ul>
                     <li id="notify-resource-id">One <code>id</code> property to identify the resource described by the description resource.</li>
-                    <li id="notify-subscription">Zero, one or many <code rel="dcterms:subject" resource="http://www.w3.org/ns/solid/notifications#subscription">subscription</code> property to identify the available <a href="#subscription-resource">subscription resources</a>.</li>
-                    <li id="notify-channel">Zero, one or many <code rel="dcterms:subject" resource="http://www.w3.org/ns/solid/notifications#channel">channel</code> property to identify the available <a href="#notification-channel">notification channels</a>.</li>
+                    <li id="notify-subscription">Zero or one <code rel="dcterms:subject" resource="http://www.w3.org/ns/solid/notifications#subscription">subscription</code> property with the property value containing zero, one or many entries to specify the available <a href="#subscription-resource">subscription resources</a>.</li>
```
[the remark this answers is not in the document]
CxRes: I see the bulk of your objections relate to changes in the data model. My goal here was to use very precise language as we are *specifying*. 

For example, in the text above, we cannot add multiple `subscription` property to the data model. We can add at most one `subscription` which contains a list of subscriptions. I do not think the original text adequately captures this.

I understand that this sort of thing makes specifications hard to read. But I am sure you agree that preciseness is the purpose of specifications. Perhaps, we need to discuss this further.

protocol.html
CxRes: This was one of the places where I should have discussed first. But I do not understand how the channel type specifies a data model. I thought the notification channel data model specifies the data model for the message. Similarly, isn't the format JSON-LD? I have not seen one type specifying these things, the only extra thing it can specify afaict is features. What am I missing here?

protocol.html
csarven: It is a practise, not a requirement. There are no "pubrules" (publication rule) that prohibits the use of different spellings - that would be very counter to the W3C Internationalization activities. Or any human checking or blocking publication because of it - as far as I know! Furthermore, the document's language is set to `en` without a subtag (which would allow even mixed spelling) - from the point of internationalisation-friendliness, it is more important that the document uses the `lang` attribute among other things, than not. See also https://validator.w3.org/i18n-checker/check?uri=https%3A%2F%2Fsolid.github.io%2Fnotifications%2Fprotocol - "No issues to report".

Having said that, editors should indeed make use of the guides available to them as much as possible and as they deem useful. This is also mentioned in Solid contributing guidelines. If no objections, I prefer to continue to use British or Canadian over American spelling, but most importantly consistently within a given document, which is why I reverted your request.

protocol.html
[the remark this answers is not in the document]
csarven: The source is https://www.w3.org/TR/security-privacy-questionnaire/#private-browsing .

I've added the `cite` attribute linking to it.

I've also went ahead and proposed the change your suggesting here at the source: https://github.com/w3ctag/security-questionnaire/pull/150 . If that is accepted, we can update our spec. Until then, let's stick to the quote as is unless we want to express it differently - I don't think we should bother at this time.

protocol.html
CxRes: Should we specify the lang as en-GB explicitly, then? The thing is, an American English inclined contributor would likely be inconsistent again in the future otherwise...

protocol.html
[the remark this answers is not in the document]
csarven: The original text indeed allows zero, one, or more subscription properties! That is, in the description resource, one can discover multiple subscription and channels to be available.

In the Turtle, that'd be expressed like:

```turtle
<>
  notify:subscription <https://websocket.example/subscription>,
                      <https://ldn.example/subscription/>,
                      <https://streaminghttp.example/subscription> .
# the above is equivalent to writing:
# notify:subscription <https://websocket.example/subscription> .
# notify:subscription <https://ldn.example/subscription/> .
# notify:subscription <https://streaminghttp.example/subscription> .

<https://websocket.example/subscription>
  notify:channelType notify:WebSocketChannel2023 .

<https://ldn.example/subscription/>
  notify:channelType notify:LDNChannel2023 .

<https://streaminghttp.example/subscription>
  notify:channelType notify:StreamingHTTPChannel2023 .
```

Again, the description need not be specific about the encoding but just focus on cardinality. It is only the example that shows how it is codified in JSON-LD.

protocol.html
CxRes: I get you but I ask that we rethink the language and make things more explicit. What confused me was the use of the word "message" which I confused with "notification message". Whereas what I understand you are trying to say is that message between Client, Resource Server and Subscription Server can change based on the notification channel type.

Re: Accept feature and message format, I think of it as orthogonal to notification channel type. I don't think we want a new channel type being defined just because you want to send the message in a different data model and format. Hence, my original objection and removal of the statement.

protocol.html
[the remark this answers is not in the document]
CxRes: I understand where you are coming from, but I have to disagree here. What you have cited as the example shows one property/predicate that has multiple values/objects. 

My contention is that we want to expressly prohibit (and make client life easier):
```json
{
  "subscription": [...],
  "subscription": [...]
} 
```

I am afraid the current langauge does not do this! I am not at all arguing that we focus on encoding. But in talking about cardinality, we must explicitly capture if we are talking about the cardinality for the predicate/property or object/property value. 

---

I guess my larger point is that I find the use of "property" and "values" in this section confusing. For example, for "features" you say:

> Zero, one or many feature **property value(s)** to state the supported features

But to follow you reasoning, we could write:

> Zero, one or many feature **properties** to state the supported features.

But obviously this is confusing, because in the notification-channel-data-model we describe feature:modifier pairs as

> Zero, one or many properties stating a particular feature

I ask that we be careful here because those using the spec may not have your knowledge. It is helpful to be slightly ignorant here; catering to that ignorance will lead to a better spec.

protocol.html
csarven: If WebSocketChannel2025 for any reason needs to require Turtle as the primary format and throw in additional core properties and possibly even features that are only meaningful/useful for WebSocketChannel2025, that'd would be allowed.

Besides those kind of scenarios, I would agree that there probably wouldn't be multiple notification channel types where their only difference would be the the data format. I can't see why WebSocketChannel2023 (sticking to JSON-LD default) and where WebSocketChannel2023-Turtle needs to be exist. This is a bit too hypothetical.

I've updated the language to indicate that it is a possibility.

csarven:
Merging this at this point as all mostly editorial changes are now in place.

Issue #152, opened 2023-02-02
Expected interactions intentionally left out of scope


elf-pavlik:
Following up on today's discussion. Some interactions we leave intentionally out of scope.
Some authentication/authorization related details are still missing, mostly between Resource Server and Subscription Server.
I recall previous version having distinct Authorization Server used by both components.

 Taking into account #142 we will have (sequence diagrams available at the very bottom of this comment):

* Content Consumer side
  * Discovery Client
  * Subscription Client
  * Notification Receiver
* Content Provider side
  * Resource Server
  * Subscription Server
  * Notification Sender

We define all interactions between the Content Consumer components and Content Provider components. All the details of interactions between components on the same side are left up to implementations.

Assumptions made about internal interactions:

#### Discovery Client & Subscription Client

Subscription Client always receives information about Subscription Resources discovered by the Discovery Client

#### Discovery Client & Notifications Receiver

Notification Receiver sometimes receives information about Notifications Channels discovered by Discovery Client

#### Subscription Client & Notifications Receiver

Notification Receiver sometimes receives information about Notifications Channels created as a result of Subscription Request made by the Subscription Client

Subscription Client always can provide *sendTo* URI in Subscription Request when request channel using it, while a Notification Receiver will receive sent notifications. Notification Receiver will receive Notification Channel information for all channels using that *sendTo*, including *sender*.

### Resource Server & Subscription Server

Resource Server has information about Subscription Resources provided by the Subscription Server which can be advertised in Description Resource

Resource Server has information about pre-established Notification Channels  which can be advertised in Description Resource (or can it come directly from Notifications Sender?)

### Resource Server & Notifications Sender

Notification Sender receives events about changes in the Resource Server which should result in sending notification on a relevant channel.

### Subscription Server & Notifications Sender

Subscription Server ensures that Notification Sender has all the information about all the Notification Channels it is responsible for sending notifications to.

```mermaid
sequenceDiagram
  participant Discovery Client
  participant Resource Server


  Discovery Client ->> Resource Server: HEAD Resource (topic)
  Resource Server ->> Discovery Client: HTTP Link rel="describedby"
  Discovery Client ->> Resource Server: GET Description Resource
  Resource Server ->> Discovery Client: Description Resource

```

```mermaid
sequenceDiagram
  participant Subscription Client
  participant Subscription Server

  Subscription Client ->> Subscription Server: Subscription Request
  Subscription Server ->> Subscription Client: Subscription Response (with Notification Channel)
```

[...]

Issue #143, opened 2023-01-10
Define all Product Classes


CxRes:
Currently, we describe product classes as building upon standards in 1.3.3, but (for some terms) we do not explicitly define what they are or explain their purpose anywhere in the document. 

Since, we link to 1.3.3 when these product classes occur in the document, it would be helpful if we defined them as well! Even if they seem self-explanatory, imho we should make them explicit.

csarven:
I'm unclear about your inquiry. Is it that some products classes are underdefined or needs better definitions, or is it to define new products?

It is true that each concept in the concept scheme (#classes-of-products) doesn't link out to the rest of the document, but the concept scheme does provide some orientation FWIW:
>These products are referenced throughout this specification.

CxRes:
Subscription Client, Subscription Server and Resource Server can do with one line definitions of what they are and/or what they do (not just what they build on) as these products are repeatedly referenced to throughout the document.

PR #162, opened 2023-03-02
extracted Discovery Client from the Subscription Client

### LINKED ISSUE #142: Rename Subscription Client
With the redesign of the Notification Protocol, "Subscription Client" is no longer appropriate nomenclature. For, our client might connect to a pre-existing notification channel without establishing a subscription at all.

I am not sure what to call it other than "Client". Perhaps "Discovery Client" might work too.

elf-pavlik:
closes solid/process#142

Changes in this PR should be taken into consideration in

* #152 
* #143


protocol.html
```diff
@@ -380,12 +380,18 @@ <h3 property="schema:name">Overview</h3>
 
                   <p>The Solid Notification protocol defines a mechanism for client applications to discover notification services available for any given resource.</p>
 
-                  <p>The following diagram shows the high-level interactions involved in discovering notification channels and subscription services. A subscription service allow Subscription Clients to request a customised notification channels.</p>
+                  <p>The following diagrams show the high-level interactions involved in discovering Notification Channels and Subscription Services. A Subscription Service allow Subscription Clients to request a customised Notification Channels.</p>
```
TallTed, line 383:
```suggestion
                  <p>The following diagrams show the high-level interactions involved in discovering Notification Channels and Subscription Services. A Subscription Service allows Subscription Clients to request a customised Notification Channel.</p>
```

TallTed (changes requested):
Small grammar fixes.

protocol.html
```diff
@@ -469,6 +473,8 @@ <h4 property="schema:name skos:prefLabel">Classes of Products</h4>
                       <span rel="skos:hasTopConcept"><span resource="#SubscriptionClient"></span><span resource="#SubscriptionServer"></span><span resource="#NotificationSender"></span><span resource="#NotificationReceiver"></span><span resource="#ResourceServer"></span></span>
 
                       <dl>
+                        <dt about="#ResourceServer" property="skos:prefLabel" rel="skos:broadMatch" resource="spec:RespondingAgent" typeof="skos:Concept"><dfn id="ResourceServer">Resource Server</dfn></dt>
+                        <dd about="#ResourceServer" property="skos:definition">A <em>resource server</em> that builds on HTTP <cite>server</cite> [<cite><a class="bibref" href="#bib-rfc7230">RFC7230</a></cite>] and [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>] by defining media types, HTTP header fields, and the behaviour of resources, as identified by link relations.</dd>
```
TallTed, line 477:
```suggestion
                        <dd about="#ResourceServer" property="skos:definition">A <em>resource server</em> builds on an HTTP <cite>server</cite> [<cite><a class="bibref" href="#bib-rfc7230">RFC7230</a></cite>] and [<cite><a class="bibref" href="#bib-rfc7231">RFC7231</a></cite>] by defining media types, HTTP header fields, and the behaviors of resources, as identified by link relations.</dd>
```

protocol.html
```diff
@@ -380,13 +380,17 @@ <h3 property="schema:name">Overview</h3>
...
 
                   <figure id="solid-notifications-flow-discovery" rel="schema:hasPart" resource="#solid-notifications-flow-discovery-subscription">
                     <img rel="schema:image" src="notifications-flow-discovery.mmd.svg" width="800" />
 
                     <figcaption property="schema:name">Solid Notifications Flow Discovery</figcaption>
                   </figure>
+                  
+                  <p>A Subscription Service allows Subscription Clients to request a customised Notification Channel.</p>
+                  
+                  <p>The following diagram shows the high-level interactions involved in establishing a customised Notification Channel.</p>
```
TallTed, line 393: Q is whether Solid docs are meant to use US or UK spelling variations. (W3C uses US.)
```suggestion
                  <p>A Subscription Service allows Subscription Clients to request a customized Notification Channel.</p>
                  
                  <p>The following diagram shows the high-level interactions involved in establishing a customized Notification Channel.</p>
```

TallTed (changes requested):
Small things...

protocol.html
csarven: The original PR remains consistent with the document. This suggestion breaks it.

We've discussed the spelling a few times in the past. The manual of style('s Spelling) is not a requirement AFAIK and have not seen it strictly applied. It is not required for Solid specs either. W3C spellchecker supports English in general, and not limited to a specific sub-dictionary.

Can we move on?

protocol.html
csarven: +1 to removing "that"
-1 to US English spelling; should be "behaviours"

TallTed: Adjusted.

protocol.html
TallTed: Given that we're operating as a [W3C CG](https://www.w3.org/groups/cg/solid/), I think we should be conforming to the [W3C Manual of Style](https://www.w3.org/Guide/manual-of-style/), and to [11.2 Spelling](https://www.w3.org/Guide/manual-of-style/#Spelling) in particular.

The [W3C Manual of Style](https://www.w3.org/Guide/manual-of-style/) would probably benefit from some attention from the [I18n](https://www.w3.org/groups/wg/i18n-core) folks, and I'd be happy to adjust my editing to such an update.

It is more than regrettable that the [W3C Manual of Style](https://www.w3.org/Guide/manual-of-style/) says (more than once!) to use US English spelling and then points to a W3C spellchecker that only supports a generic English which appears to accept both US and UK spellings, but at least that checker only works *on w3.org URIs*.  (It doesn't do anything with GitHub pages, nor within my local browser or other editor, where I am set up to use US English, as advised.)  The W3C spellchecker is thoroughly problematic, as it can easily lead to one document that contains a mix of spellings _of the same word_, which is only confusing.

If the Solid Project is to have a different Style than W3C, I'd like that to be codified somewhere, so there's no question of whether spellings should be `ou` or `o`, or `s` or `z`, etc.

You are of course free to reject this requested/suggested change. I would suggest, in that case, that an issue should be logged on the specification repo, toward a definitive conclusion going forward.

elf-pavlik: tracked in https://github.com/solid/process/issues/318 which can be transferred to another repo if needed

csarven (approved):
Let's leave out changes that are not pertinent to this PR for issues or other PRs, e.g., reordering classes of products :)

PR #174, opened 2023-05-03
Change JSON-LD context URI in Solid Notifications Protocol

csarven:
As per [Solid CG 2023-05-03 meeting](https://github.com/solid/specification/blob/d9c7d62b32aef199b3b84f29840bba93bd6fe5e9/meetings/2023-05-03.md#solid-notifications-vocab-and-json-ld-context) decision with @timbl (as W3C Team contact), we'll need to change the JSON-LD context URL to something like `ns/solid/notifications-context/v1`.

This is a "Corrections that do not add new features" as per https://www.w3.org/2021/Process-20211102/#class-3 , in particular:

>makes conforming data, processors, or other conforming agents become non-conforming according to the new version

This entails that the change here in the Editor's Draft is towards the next release of the Notifications Protocol (0.3.0).

Since the JSON-LD context was never published, I don't suspect any implementation depending on it in the sense that they're fetching it. Besides, the specification only required that implementations recognise the URI, and the reject requests of what's unfamiliar. So, current IP implementations will need to change the URI. I suspect that this won't be a major issue as there will be other changes coming in with 0.3.0.

I suggest that all Notification Channel Types mentioning the JSON-LD context follow this PR.

If anyone has any strong opinions on what the URI should be, **NOW IS THE TIME**.

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -515,7 +515,7 @@ content: "";
               <p>A Resource Server may provide notifications over a notification channel that might be of one of many <a href="#notification-channel-types">notification channel types</a> that build upon existing web protocols.
               Individual <a href="#notification-channel-types">notification channel types</a> may augment the discovery and subscription requirements of the notification protocol.</p>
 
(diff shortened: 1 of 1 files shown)


elf-pavlik (approved):

csarven:
Merged as per agreement in 2023-05-18 meeting: https://github.com/solid/notifications-panel/blob/1de1b81f78a75a5dd3339ca9acea37d7f1e5072e/meetings/2023-05-18.md#change-json-ld-context-uri-in-solid-notifications-protocol

Issue #159, opened 2023-02-23
Filter feature


csarven:
Follows https://github.com/solid/notifications-panel/blob/main/meetings/2023-02-02.md#add-ldn-channel-2023-ed to introduce a feature to allow fine-grained notification filtering, e.g., activity types, activities involving particular actors, tags.

It is based on a general requirement as suggested in https://github.com/solid/notifications-panel/issues/1#issue-489650894 :

>Mechanism for actors to make a request to be notified about particular activities.

An example model is mentioned/proposed in https://github.com/solid/notifications/pull/147#issuecomment-1410518467 , where a `filter` property would have an object detailing the information to filter on:

```json
{
    "filter": {
       "type": "Add",
       "actor" "https://csarven.ca/#i"
    }
}
```

And it is an easy jump to see how something like this can work too (if I'm not messing up the syntax here -- simplified here without JSON-LD context):

```json
{
    "filter": {
       "id": {
         "type": "oa:Annotation",
         "oa:hasTarget": "https://csarven.ca/archives/linked-research-decentralised-web/ce36de40-64a7-4d57-a189-f47c364daa74#abstract"
         "dcterms:rights": "https://creativecommons.org/licenses/by/4.0/",
       },
       "type": "Add",
       "actor": "https://csarven.ca/#i"
    }
}
```

CxRes:
This is a general point about features: Could we move them out of the spec or at least to an appendix? This way we can release them independently of the main spec (and not tie it to milestones).

csarven:
A milestone generally includes all kinds of new features, and so tied in that respect to a release.

I think it'd be important to keep the core `feature`s in the Notifications Protocol because they'd need to be supported by all implementations. Similar to e.g., subscription/channel data models. And features that are unique to notification-channel-types can be in their respective specifications, and so released independently. This was the approach for filter/activityType which originally proposed for LDNChannel2023, but then deemed to be useful for all notification-channel-types.

Having said that, whether a particular feature (for example `filter`) should be in 0.3 (or whenever ready) is something we can be flexible and decide.

damooo:
Though simple filter object works in most cases, and are sufficient, it may be good to consider `acp` like matcher+policy mechanism.

That would allow us to express more general filter(-or-for-any-other) policy over a context composed of activity info, object info, etc. It may allow any-of/all-of/none-of combinatrics of simple match graphs. But whether such general and powerful mechanism is required is a question.

damooo:
[...]

PR #176, opened 2023-05-15
Add notify:filter feature

csarven:
Filter feature to refine the characteristics of notifications (towards resolving https://github.com/solid/notifications/issues/159 ).

I suggest that we work on modeling the core filter object (in subsequent PRs), e.g., `type` is a good candidate to filter activity types, and possibly `actor` .. so perhaps the object can be close to an AS2 Object (including Activity), but it should be possible to extend further e.g., Add activities of Annotation objects with a particular license. Makes sense? We can continue in 159.


elf-pavlik:
Is it really useful as defined in this PR? 

> The proposed characteristics of a notification to be sent to receiver represented as an object.

I think we need to be more specific. If we want to allow matching based on any triple pattern we can most likely specify that right away.

csarven:
Sure, let's have rough consensus in https://github.com/solid/notifications/issues/159

PR #182, opened 2023-07-06
explicit channel auth scheme

### LINKED ISSUE #155: Make AuthN/AuthZ of each channel explicit
Currently each channel doesn't specify AuthN/AuthZ method which it relies on. Instead we have implicit relying on what Channel type defines. I think each channel should specify what AuthN/AuthZ the *receiver* or the *sender* should use when receiving or sending notifications.

I'm going to track this issue inline in:
* WebHookChannel2023 #146 
* StramingHTTPChannel2023 based on discussion in https://github.com/CommunitySolidServer/CommunitySolidServer/issues/1574

Related issues:

* #134 
* #148 
* #18

elf-pavlik:
closes #155

### TODO

* [ ] Once agreed on, update examples in all the channel types

---

[Preview](http://htmlpreview.github.io/?https://github.com/solid/notifications/blob/6012fc8b4d88aab8638b94b520c1aacb6b07e4fe/protocol.html) | [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fsolid.github.io%2Fnotifications%2Fprotocol&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fsolid%2Fnotifications%2F6012fc8b4d88aab8638b94b520c1aacb6b07e4fe%2Fprotocol.html) (doesn't show the changes in code snippets!)

### DIFF
--- a/protocol.html
+++ b/protocol.html
@@ -755,6 +755,12 @@ tfoot dd + dt { margin-top:0; }
 <code>  "subscription": [{</code>
 <code>    "id": "https://websocket.example/subscription",</code>
 <code>    "channelType": "WebSocketChannel2023",</code>
+<code>    "channelAuthScheme": ["CapabilityURL"],</code>
+<code>    "feature": ["endAt", "rate", "state"]</code>
+<code>  }, {</code>
+<code>    "id": "https://webhook.example/subscription",</code>
+<code>    "channelType": "WebhookChannel2023",</code>
+<code>    "channelAuthScheme": ["CapabilityURL", "SolidOIDC"]</code>
 <code>    "feature": ["endAt", "rate", "state"]</code>
 <code>  }],</code>
 <code>  "channel": [{</code>
@@ -762,6 +768,7 @@ tfoot dd + dt { margin-top:0; }
 <code>    "type": "WebSocketChannel2023",</code>
 <code>    "topic": "https://example.org/guinan/profile",</code>
 <code>    "receiveFrom": "wss://websocket.example/guinan/profile",</code>
+<code>    "authScheme": "None",</code>
 <code>    "rate": "PT1M"</code>
 <code>  }]</code>
 <code>}</code></pre>
@@ -779,6 +786,7 @@ tfoot dd + dt { margin-top:0; }
                   <ul>
                     <li id="notify-subscription-id">One <code>id</code> property to identify the subscription service.</li>
                     <li id="notify-channelType">One <code>channelType</code> property to state the <a href="#notification-channel-types" rel="rdfs:seeAlso">notification channel type</a>.</li>
+                    <li id="notify-channelAuthScheme">One or many <code>channelAuthScheme</code> property to state the supported authentication schemes.</li>
                     <li id="notify-feature">Zero, one, or many <code>feature</code> properties to state the supported <a href="#notification-features" rel="rdfs:seeAlso">features</a>.</li>
                   </ul>
 
@@ -791,6 +799,7 @@ tfoot dd + dt { margin-top:0; }
 <code>  ],</code>
 <code>  "id": "https://websocket.example/subscription",</code>
 <code>  "channelType": "WebSocketChannel2023",</code>
+<code>  "channelAuthScheme": ["CapabilityURL"],</code>
 <code>  "feature": ["endAt", "rate", "state"]</code>
 <code>}</code></pre>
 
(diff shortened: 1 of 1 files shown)


CxRes:
I would suggest that you add an implementation guidance box, otherwise spec users will wonder why we are adding channel auth in the examples even though it has no mention in the spec.

CxRes:
That should be "zero, one or many" (I disagree with @csarvens way of describing, we have argued this previously, because there is no way to indicate that Subscription Service may not advertise this property at all and that would be OK). I would also suggest the property on Subscription Service be called `authSchemes` instead of `channelAuthScheme` for consistency with Notification Channels.

elf-pavlik:
>  I would also suggest the property on Subscription Service be called `authSchemes` instead of `channelAuthScheme` for consistency with Notification Channels.

I think `chanelAuthScheme` is consistent with `channelType`, also having two different predicates only differ by final `s` could be confusing.

> That should be "zero, one, or many"

If we want it always to be explicit it can't be zero, one of the examples uses `None`. In what scenarios would you see zero?

CxRes:
> I think chanelAuthScheme is consistent with channelType, also having two different predicates only differ by final s could be confusing.

Plural indicates that a thing is an array. Not having an "s" is the inconsistent thing, imho. 

I strongly disagree with your definition of consistency. `AuthSchemes` are only meaningful in the context of the channel, but is a general thing. `channelType` is called so because `type` is vague thing even when described in the context of a channel (and can be confused with other kinds of "types" like "data type") and that we explicitly define a "Channel Type" in the spec. 

> If we want it always to be explicit it can't be zero, one of the examples uses None. In what scenarios would you see zero?

You could conceivably have a channel that you need to subscribe to (one a server will create on request only), but is otherwise public and does not require authorization (not that I would recommend anyone make one, but consistency demands that we allow such a thing).

What I would concede is that if you specify the property, then zero entries makes no sense. But you should be able to skip specifying `channelAuthScheme` altogether in the above scenario.

protocol.html
```diff
@@ -807,6 +816,7 @@ <h3 property="schema:name" content="Notification Channel Data Model">Notificatio
                   <ul>
                     <li id="notify-channel-id">One <code>id</code> property to identify the notification channel.</li>
                     <li id="notify-channel-type">One <code>type</code> property to state the <a href="#notification-channel-types" rel="rdfs:seeAlso">notification channel type</a>.</li>
+                    <li id="notify-channel-authScheme">One <code>authScheme</code> property to state the authentication scheme used for the channel</a>.</li>
```
csarven, line 819:
```suggestion
                    <li id="notify-channel-authScheme">One <code>authScheme</code> property to state the authentication scheme used for the channel.</li>
```

protocol.html
```diff
@@ -779,6 +786,7 @@ <h3 property="schema:name" content="Subscription Service Data Model">Subscriptio
                   <ul>
                     <li id="notify-subscription-id">One <code>id</code> property to identify the subscription service.</li>
                     <li id="notify-channelType">One <code>channelType</code> property to state the <a href="#notification-channel-types" rel="rdfs:seeAlso">notification channel type</a>.</li>
+                    <li id="notify-channelAuthScheme">One or many <code>channelAuthScheme</code> property to state the supported authentication schemes.</li>
```
csarven, line 789: I couldn't suggest a change to add a paragraph/section (perhaps right after #authentication-authorization or a new subsection of the #protocol section.) The following suggestion would refer to Authentication Scheme.

```suggestion
                    <li id="notify-channelAuthScheme">One or many <code>channelAuthScheme</code> property to state the supported <a href="#authentication-scheme">authentication schemes</a>.</li>
```

protocol.html
```diff
@@ -807,6 +816,7 @@ <h3 property="schema:name" content="Notification Channel Data Model">Notificatio
                   <ul>
                     <li id="notify-channel-id">One <code>id</code> property to identify the notification channel.</li>
                     <li id="notify-channel-type">One <code>type</code> property to state the <a href="#notification-channel-types" rel="rdfs:seeAlso">notification channel type</a>.</li>
+                    <li id="notify-channel-authScheme">One <code>authScheme</code> property to state the authentication scheme used for the channel.</li>
```
csarven, line 819: The #authentication-scheme section should describe that the absence of this property is required (or not). If present, it would entail that the schemes are known, including a scheme value such as "None" meaning that authentication is not required.

I presume "None" will map to `notify:None` .. but we mean something like `notify:AuthenticationSchemeIsNotUsed`?

csarven (approved):
Thanks for putting this together and thinking about this problem/solution. Pardon the late reply.

Approving but see change requests. I'd like to see a short paragraph/section for #authentication-scheme describing a bit further. See review comments.

csarven:
I think requiring `channelAuthScheme` with at least one value encourages (more) complete and useful information. Just as specifying the type of the channel, its endpoint, features. Why wouldn't authentication scheme information be part of this? The only thing I can think of is that it is unknown (unprovided) to the DescriptionResource and that the values may change (for some reason).

As for singular/plural, singular suffices for RDF properties as atomic statements / arcs in graph.

PR #186, opened 2023-08-03
[WebSocketChannel2023] update example

elf-pavlik:
use onclose and onmessage as property not method

### DIFF
--- a/websocket-channel-2023.html
+++ b/websocket-channel-2023.html
@@ -433,8 +433,8 @@ content: "";
                   <figure id="websocket-onclose-onmessage" class="example listing" rel="schema:hasPart" resource="#websocket-onclose-onmessage">
                     <p class="example-h"><span>Example</span>: Handling WebSocket events after the connection is established.</p>
 
-                    <pre about="#websocket-onclose-onmessage" property="schema:description" typeof="fabio:Script"><code>ws.onclose(evt => reconnect(evt));</code>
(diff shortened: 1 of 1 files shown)


websocket-channel-2023.html
```diff
@@ -423,7 +423,7 @@ <h3 property="schema:name">Subscription Example</h3>
                   <figure id="websocket-new" class="example listing" rel="schema:hasPart" resource="#websocket-new">
                     <p class="example-h"><span>Example</span>: Establishing a new WebSocket connection.</p>
 
-                    <pre about="#websocket-new" property="schema:description" typeof="fabio:Script"><code>const ws = new WebSocket("wss://websocket.example/?auth=Ys3KiUq", "http://www.w3.org/ns/solid/notifications#WebSocketChannel2023");</code></pre>
+                    <pre about="#websocket-new" property="schema:description" typeof="fabio:Script"><code>const ws = new WebSocket("wss://websocket.example/?auth=Ys3KiUq");</code></pre>
```
csarven, line 426: Can you please quick remind me why is this no longer necessary?

elf-pavlik: I'll double-check it. When I used it with channel type URI as the protocol it was resulting in some error. Simply removing it worked. BTW I think this is only used in a non-normative example, if we want the channel type to rely on it would need to be specified in a normative way.

